Issue #11, opened 2018-09-09
Rename macOS app to VSCodium.app


MatiasConTilde:
As pointed out in https://github.com/Homebrew/homebrew-cask/pull/51727#discussion_r216155232, the token has to be named the same way as the App file to be included in Homebrew Cask, this means either renaming the app file to VSCodium.app or changing the token to `code-oss`. I personally like the first option more as it includes the name of this project so it isn't that general.

stripedpajamas:
@MatiasConTilde first of all thank you for spearheading the Brew issue. I'm for renaming the app to VSCodium as well, as that is a very small change pre-build and doesn't affect the end product at all (besides aesthetically). Instead of just renaming the macOS app, I think it would be more straightforward to edit the `product.json` file before building. This would affect the name of the resulting app for all platforms. 

My thought would be to leave the data directory name (`.vscode-oss`) untouched, so that the config/extensions used in anyone's built-from-source copy would be shared in a downloaded copy of VSCodium.

Issue #9, opened 2018-08-30
Add brew cask


MatiasConTilde:

stripedpajamas:
I think this would be great! I don't know much about making brew casks. Because this is not "official" it would have to go into the unofficial casks, right?

MatiasConTilde:
I already went ahead and submitted it myself (https://github.com/Homebrew/homebrew-cask/pull/51727), it wasn't that difficult after all :)

According to the [Finding a Home For Your Cask](https://github.com/Homebrew/homebrew-cask/blob/master/doc/development/adding_a_cask.md#finding-a-home-for-your-cask) guide, an Unofficial Cask is
> An allegedly unmodified compiled binary, by a third-party, of a binary that has no existing build by the owner of the source code.

As VSCodium is kind of a fork (so not unmodified), and Microsoft (the owner of the source code) doesn't already build it, I think that it counts as official, but we'll have to wait for the maintainers approval to confirm.

vitorgalvao:
HBC maintainer (and writer of that rule) here.

That rule only applies where the name could lead to confusion. For example, if MS didn’t provide a compiled app but you did **and called it Visual Studio Code**, it would fall under that rule. Basically: if by name alone a user could think they’re installing something official when it’s not, the rule comes into play. As the name is different, the issue doesn’t come up.

stripedpajamas:
`brew cask install vscodium` works now 🎉 

Thanks for your help @MatiasConTilde (and @vitorgalvao). I will update the readme.

PR #12, opened 2018-09-10
Rename app to VSCodium

stripedpajamas:
Also set name of release to the tag name to avoid any unrelated commit messages from muddying that up.


stripedpajamas:
#11 #9

PR #19, opened 2018-09-19
If vscode folder exists, fetch updates not invoke clone

huksley:


huksley:
I will locally rebuild when new version will be available.
I suppose other people will do the same, without using your binaries.
It is actually works wonderfully, thanks @stripedpajamas!

You`ve been mentioned here:
https://lobste.rs/s/2lepmm/you_think_visual_studio_code_binary_you

stripedpajamas:
Oh I see. I don't see any benefit of this repo if you're building from source yourself... Doing a git pull on Microsoft's repo, checking out the latest tag, and running the one os-specific gulp build task would be a much smaller and simpler shell script than all the stuff we had to cobble together to get Travis doing it. 

I'll go ahead and merge anyway since it doesn't hurt anything 👍

PR #25, opened 2018-09-30
Better oss icon

### LINKED ISSUE #4: Better Icon
We should come up with a better icon for the FOSS version. The FOSS icon is really ugly.  That would be sugar on top of the cake. 👍

clarsonneur:
Fix #4


clarsonneur:
I need to find a way to create icns files... The online converter seems to fail... But I should be able to do it with gimp.

PR #27, opened 2018-10-02
Undo telemetry before build

### LINKED ISSUE #26: VSCodium still accessing MS data gathering URLs
Running the latest build of VScodium on maxOS High Sierra.

As reported by the Little Snitch firewall. The binary is trying to access telemetry domains:

`dc.services.visualstudio.com` 
`dc.trafficmanager.net`
`vortex.data.microsoft.com`

These are MS data gathering URLs. If one of the intentions of this project is to disable Telemetry it would be good to remove the communication to these.

There has been quite a bit of discussion around this on the VScode github page:

https://github.com/Microsoft/vscode/issues/16131

And on Reddit:

https://www.reddit.com/r/privacy/comments/80d8wu/just_realised_that_visual_studio_code_sends/

Telemetry in VSCode can be turned off, but not fully. Instead of disabling it altogether, they just register that you have it turned off, the server communication still happens. 

The URL `https://vortex.data.microsoft.com/collect/v1` is referenced in the VSCode source, so it could be replaced with something innocuous via Sed, that doesn't stop the other domains or any future ones they add. The others are likely added via the code's npm/build dependencies. 

So it may be best to just simply add information on the "issue" and instructions to this projects README.MD to work around the problem by utilising your local hosts file to block the domains.

eg. Adding the line to your `/etc/hosts` file:

`0.0.0.0 dc.services.visualstudio.com dc.trafficmanager.net vortex.data.microsoft.com`

I'm interested in hearing what others think about this.

stripedpajamas:
Fix #26 

Replaces telemetry endpoints with 0.0.0.0. I tested that this does not mess up the build, and [Little Snitch](https://www.obdev.at/products/littlesnitch/index.html) shows no outgoing traffic from the binary after opening (unless I go to the Marketplace, but that is expected behavior).


em92:
How about using patch files instead of sed, xargs etc?

stripedpajamas:
@em92 I would love to not use `sed` if it makes sense to... what would using patch files look like? 

The reason I used `sed` and `grep` is because there are about 12 files that have the link scattered through different dependencies' dependencies.

JL2210:
You could use `sed` and then create a patch file from that. Simply store the patch created by `diff -Naur` in this repository.

stripedpajamas:
@em92 @JL2210 thanks for the patch tip, I hadn't used diff/patch in this way before. I've updated the PR, if you have any objections let me know.

JL2210:
Not exactly what I would do. Usually, I would rename source tree to `a`, and the source tree with modifications to `b`. Then I would apply it with `patch -Np1 -i <patch>`.

JL2210:
How did you make a patch like this? I've never been able to do that.

stripedpajamas:
@JL2210 probably could have been done a better way but:

- cloned vscode, installed dependencies
- copied to vscode2, cd'd there
- ran `grep | sed` commands, `cd ..`
- ran `diff -Naur vscode vscode2 > telemetry.patch`
- went through patch file, changed paths to be relative to inside the vscode directory
- tested

stripedpajamas:
After @paulcarroty commented in #26 about using a regex I started wondering why we would go the patch route. Using the regex means that if MS adds another dependency that hits either of these URLs, we will still end up stripping it out. Using the patch file means we would have to a) notice the addition, and b) update the patch file. Thoughts @em92 @JL2210 ? I am inclined to revert back to `grep | sed`

JL2210:
That might be a good idea, though put the `grep | sed` commands in either a file called `patch.sh` or `undo-telemetry.sh`.

em92:
> After @paulcarroty commented in #26 about using a regex I started wondering why we would go the patch route

Actually I meant using patch files to cut out telemetry modules, not just zero addresses. If zero addresses, grep+sed is good enough.

PR #32, opened 2018-10-04
32 bit builds

stripedpajamas:
Thanks to @JL2210 for starting this initiative. Ended up running the 32-bit builds in a separate linux environment and using an env var to determine certain things.

changes:
- set the `npm_config_arch` env var to `ia32` before running `yarn` if building 32-bit
- install 32-bit dependencies (libc6-dev-i386, gcc-multilib, g++-multilib, libx11-dev:i386, libxkbfile-dev:i386) if building 32-bit
- adjust the check_tags logic to account for 64-bit vs 32-bit built linux files
- move the setting of the current tag -- there was a bug where builds would be deployed as untagged and this should fix that
- update README for 32-bit support
- stop using `npx` as the builds have a chance of overflowing the heap -- using MS's `npm run gulp` which sets a larger heap size during builds

built files are [here](https://github.com/stripedpajamas/vscodium/releases/tag/1.27.2) if anyone wants to test before merging

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ matrix:
   include:
     - os: linux
       sudo: required
+    - os: linux
+      sudo: required
+      env: BUILDARCH=32
     - os: osx
 
 language: node_js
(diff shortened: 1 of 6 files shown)


JL2210:
Try x86 instead of x32. That's pretty much my only suggestion.

JL2210:
Nice.

Issue #41, opened 2018-10-12
Auto Updating Support


ganapativs:
How does auto updating works here? how are we pushing the new update?

also anyone has idea about how to configure custom `update.channel`?

JL2210:
Use the deb repository. See the README.

stripedpajamas:
The deb repository might be a good idea for Linux users. For Mac users we will have to figure something else out. It's not the best experience having to check the repo for updates or ask brew for updates. This could use some discussion.

JL2210:
Brew can auto-update. In Debian, auto-updates do not happen automatically. You still have to type `sudo apt-get update && sudo apt-get upgrade`.

stripedpajamas:
I think we can get something like that working. I need to investigate out it currently works in vscode

ganapativs:
Yes, there must be a proper interface for that feature, we need to check how to configure the same. I'll see if I can figure out something.

tynx:
In case anyone want's to do it with a simple bash-"script", that you can trigger as often as you like...
```bash
wget -q -O /tmp/vscodium.deb $(wget -O- -q https://api.github.com/repos/VSCodium/vscodium/releases/latest | grep browser_download_url | cut -d '"' -f 4 | grep "amd64.deb") && sudo apt -f install /tmp/vscodium.deb && rm /tmp/vscodium.deb
```
Note: This Only supports dpkg/apt and amd64... :-)

tynx:
Out of personal interesst I looked into the source of the updater-mechanism...

https://github.com/Microsoft/vscode/blob/bb23217d1ef48ff6549b214162b36e8b8306e943/build/gulpfile.vscode.linux.js#L84

It seems that the update mechanisms hits this URL. See the following folder for the concrete implementation of the updater-service. the abstract implementation contains the basic URL-schemantics for all platforms and the actual "update-checking" happens in the platform-specific implementations.
https://github.com/Microsoft/vscode/tree/master/src/vs/platform/update/electron-main

for linux:
https://github.com/Microsoft/vscode/blob/ab63c5d4c98a7deaf4974e8a8723f8dd5d7278b9/src/vs/platform/update/electron-main/updateService.linux.ts#L39

I'd probably try to change the update-url in the gulp-build-process to localhost or similar and would see what gets requested and try to respond... Maybe that insight helps - if not just ignore my input :-)

ganapativs:
[...]

PR #47, opened 2018-10-15
add basic structure and explanation for fake api

tynx:
each subfolder should/will be populated with version files...

this is a basic draft based on #41

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -1,0 +1,8 @@
+# Faking the update-API of Microsoft
+
+There is a distinction between stable and insider for the "quality" of the version. Also there are three platforms: linux, win32 and darwin. For each version there is a subfolder. There should be a file in each of the folders called "VERSION" (uppercase).
+
+By example of the api of MS itself (darwin/stable):
+https://vscode-update.azurewebsites.net/api/update/darwin/stable/VERSION
+
+The repo containing the version-json-files should have pages activated, see project settings. Instructions here: https://pages.github.com/


stripedpajamas:
I am thinking that we will only support the `stable` release, as at least currently the whole build process works by looking at the latest tag on the MS repo. So I would say remove the insider folders from this PR. 

Also, I don't think we will need to host the repo on Github pages. It should be that if the file is there (and updated at build time), then VSCodium could access it directly at https://github.com/vscodium/vscodium/master/api/update/{os}/stable/VERSION. So that part can be removed from the readme.

api/README.md
```diff
@@ -0,0 +1,8 @@
+# Faking the update-API of Microsoft
+
+There is a distinction between stable and insider for the "quality" of the version. Also there are three platforms: linux, win32 and darwin. For each version there is a subfolder. There should be a file in each of the folders called "VERSION" (uppercase).
+
+By example of the api of MS itself (darwin/stable):
+https://vscode-update.azurewebsites.net/api/update/darwin/stable/VERSION
+
+The repo containing the version-json-files should have pages activated, see project settings. Instructions here: https://pages.github.com/
```
stripedpajamas: This line can be removed (shouldn't need to use Github pages)
```suggestion
```

api/README.md
```diff
@@ -0,0 +1,8 @@
+# Faking the update-API of Microsoft
+
+There is a distinction between stable and insider for the "quality" of the version. Also there are three platforms: linux, win32 and darwin. For each version there is a subfolder. There should be a file in each of the folders called "VERSION" (uppercase).
```
stripedpajamas:
```suggestion
There is a distinction between stable and insider for the "quality" of the version. Also there are three platforms: linux, win32 and darwin. For each platform there is a subfolder for the quality. There should be a file in each of the folders called "VERSION" (uppercase).
```

JL2210:
@stripedpajamas You can commit directly to his repository if he's not responding.

stripedpajamas:
PR seems to be abandoned and more work needs to go into signing the build output before any updating API can be considered.

PR #49, opened 2018-10-15
adding missing dependency jq

tynx:
While using the script I realized that the package jq was missing


JL2210:
Question: What is `jq`?

PR #55, opened 2018-10-30
Fix big darwin file

### LINKED ISSUE #54: VSCodium is more than twice the size of VSCode on MacOS
Hello,
I have been following vscodium for a couple of months now, and decided to give it a try, but…
schermata 2018-10-29 alle 23 14 48
In the screen both of them were just downloaded from [vscodium](https://github.com/VSCodium/vscodium/releases/download/1.28.2/VSCode-darwin-1.28.2.zip) and [vscode](https://az764295.vo.msecnd.net/stable/7f3ce96ff4729c91352ae6def877e59c561f4850/VSCode-darwin-stable.zip).

I'm no expert of build pipelines, so I don't know if this is expected or not, but surely 430MB of size seems odd. Sorry if I get it wrong.

I also inspected both packages and inside VSCodium it appears to be three Electron.Framework bundles (by 70MB each), while in VSCode there is only one.

stripedpajamas:
| Package        | Size           |
| ------------- |-------------|
| VSCode@1.28.2      | __192.8mb__ |
| old VSCodium@1.28.2     | __428mb__      |
| new VSCodium@1.28.2 | __189.3mb__      |

Closes #54

### DIFF
--- a/create_zip.sh
+++ b/create_zip.sh
@@ -3,7 +3,7 @@
 if [[ "$SHOULD_BUILD" == "yes" ]]; then
   if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
     cd VSCode-darwin
-    zip -r ../VSCode-darwin-${LATEST_MS_TAG}.zip ./*
+    zip -r -X -y ../VSCode-darwin-${LATEST_MS_TAG}.zip ./*
   elif [[ "$BUILDARCH" == "ia32" ]]; then
     cd VSCode-linux-ia32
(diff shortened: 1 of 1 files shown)


stripedpajamas:
I think to avoid an extraneous homebrew PR, I will wait until the November release of vscode to merge this.

JL2210:
Try not to strip file attributes. They could be important.

stripedpajamas:
`-X -y` is how MS is doing it [here](https://github.com/Microsoft/vscode/blob/master/build/tfs/darwin/product-build-darwin.yml), so there is no reason for us not to.

PR #56, opened 2018-11-04
Fix arm64 build

JL2210:
(Hopefully!) Fixes arm64 build for Travis and improves the ia32 build.

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -19,7 +19,7 @@ else
   elif [[ $BUILDARCH == "arm64" ]]; then
     sudo dpkg --add-architecture arm64
     sudo apt-get update
-    sudo apt-get install libc6-dev:arm64 gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
+    sudo apt-get install libc6-dev-arm64-cross gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
     sudo apt-get install libx11-dev:arm64 libxkbfile-dev:arm64
     export CC=/usr/bin/aarch64-linux-gnu-gcc
     export CXX=/usr/bin/aarch64-linux-gnu-g++
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -12,10 +12,18 @@ else
     sudo apt-get update
     sudo apt-get install libc6-dev-i386 gcc-multilib g++-multilib
     sudo apt-get install libx11-dev:i386 libxkbfile-dev:i386
+    export CC="/usr/bin/gcc -m32"
+    export CXX="/usr/bin/g++ -m32"
+    export CC_host=/usr/bin/gcc
+    export CXX_host=/usr/bin/g++
   elif [[ $BUILDARCH == "arm64" ]]; then
-    # Use the default C / C++ compilers,
-    # because some makefiles default to CC:=gcc:
-    export CC=/usr/bin/cc
-    export CXX=/usr/bin/c++
+    sudo dpkg --add-architecture arm64
+    sudo apt-get update
+    sudo apt-get install libc6-dev:arm64 gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
+    sudo apt-get install libx11-dev:arm64 libxkbfile-dev:arm64
+    export CC=/usr/bin/aarch64-linux-gnu-gcc
+    export CXX=/usr/bin/aarch64-linux-gnu-g++
+    export CC_host=/usr/bin/gcc
+    export CXX_host=/usr/bin/g++
   fi
 fi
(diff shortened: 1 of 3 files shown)


install_deps.sh
```diff
@@ -12,10 +12,18 @@ else
     sudo apt-get update
     sudo apt-get install libc6-dev-i386 gcc-multilib g++-multilib
     sudo apt-get install libx11-dev:i386 libxkbfile-dev:i386
+    export CC="/usr/bin/gcc -m32"
```
stripedpajamas: Are these 32-bit changes necessary in this PR? I'm wary to change these if there are no open issues about the 32-bit binaries being broken.

stripedpajamas:
Can we test the binaries before merging? If you add a `GITHUB_TOKEN` env var to your Travis build with a personal access token (https://github.com/settings/tokens/new), Travis should build+upload to your forked releases (https://github.com/JL2210/vscodium/releases). That way @435vic could try it out (if you haven't confirmed yourself already).

install_deps.sh
JL2210: I don't really want to be worried by the fact that there may be x86_64 code in the 32-bit versions. This way, we can be sure.

JL2210:
@stripedpajamas How do I do that? I'm a bit wary about letting Travis have delete access for my GitHub repository.

435vic:
Don't worry, I've merged your `fix-arm64-build` branch in to my fork, and it's building now. Check it out [here](https://travis-ci.com/435vic/vscodium).

435vic:
Just tried it, and it still doesn't work.
`pty.node: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=5eda743055b93473214eedeb4b4c7faea6952838, not stripped`

435vic:
Just so you know, I built VSCode on my arm64 machine and the output is different:
`pty.node: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, BuildID[sha1]=b2994f46672b0de2d727d7e47caf2487953abb67, not stripped
`
As you can see, the node modules _are_ built as arm64.

435vic:
Just for the sake of trying it out, could you also set `npm_config_target_arch` to arm64? On [the issue](https://github.com/Microsoft/vscode/issues/61623#issuecomment-435162199) in official repository, headmelted said it might be a good idea to do so.

Also, the build logs on Travis say that `libx11-dev:arm64` and `libxkfile-dev:arm64` cannot be found.

JL2210:
@435vic Check my new commit.

install_deps.sh
stripedpajamas: That's understandable. For safety, please move these lines to a different PR if you think they are necessary and test the resulting binaries on a 32-bit machine to confirm nothing is broken.

stripedpajamas:
@435vic are you able to check the binaries of @JL2210's latest commit?

435vic:
Sorry for taking so long to respond, I'm building the PR now.
Just noticed something, it seems the newest build scripts for VSCode seem to now check the yarn version, which is actually very, very, old.
image
You should update the yarn version, check the instructions on the yarn website [here](https://yarnpkg.com/en/docs/install-ci).

JL2210:
Isn't that what I suggested?

stripedpajamas:
Yes but the latest code on VSCodium master installs the latest yarn version. @JL2210 might not have pulled master into the arm branch lately. Once that's done it will build with latest yarn.

JL2210:
@435vic @stripedpajamas Anything new on this?

JL2210:
Sure.

JL2210:
I will.

JL2210:
@stripedpajamas Done.

stripedpajamas:
@435vic can you try the binaries again?

stripedpajamas:
@JL2210 can you test this yourself (do you have access to an arm64 environment)? You will need to create a temporary personal access token with `public_repo` permissions and add it as an environment variable in the Travis build called `GITHUB_TOKEN`. If you are uncomfortable doing this, we can continue to wait for @435vic.

JL2210:
@stripedpajamas Sorry. I don't have any arm64 computers.

JL2210:
And both of my laptops are not powerful enough to run VMs.

stripedpajamas:
@JL2210 was thinking of merging this in as it certainly can't make things worse (since arm64 builds don't work and this still might fix them, pending more testing).

JL2210:
I'm certainly not against it.

JL2210:
I'm certainly not against it.

PR #62, opened 2018-12-03
Only move artifacts if we built

stripedpajamas:


stripedpajamas:
@pavanagrawal123 does this look good to you? I have it skipping the `mv` and `PublishArtifact` stuff if `SHOULD_BUILD` is not `yes`. Just want to sanity check before merging.

pavanagrawal123:
Yep looks good.

Just FYI running a build on PR checks is very dangerous because the way release pipeline is configured.

stripedpajamas:
What can we do to make it safer? It is safe on Travis because the github token is not present in PR builds.

pavanagrawal123:
A release will happen automatically if the build pipeline succeeds. So, if someone were to PR a change that compromises the integrity of the package, the PR would be built and automatically deployed to GH if should build is true. IMO there's no need to do PR checks for Azure Pipelines, if the build works in Travis, it will work in Azure.

stripedpajamas:
It looks like SHOULD_BUILD can't be yes if there is no GITHUB_TOKEN variable (https://github.com/VSCodium/vscodium/blob/master/check_tags.ps1#L10) and now that I've marked GITHUB_TOKEN as secret, it won't be included in any PR builds. So that should keep any publishing from happening, right? I'd like to keep Travis + Azure running on PRs if possible for consistency sake.

pavanagrawal123:
yep looks good :)

Issue #66, opened 2018-12-12
32-bit Windows packages.


JL2210:
I believe that Windows and Wine users will be happy if we provide 32-bit Windows packages. Any suggestions?

pavanagrawal123:
I'd be willing to work on this if wanted :)

JL2210:
Thank you. That would be much appreciated.

Xandernardo:
Just what I was looking forwards too, as in my work I only have access to a Win7 x86 machine

JL2210:
@pavanagrawal123 Have you started work yet?

Thilas:
@pavanagrawal123, any progress on a 32-bit release? Thanks!

stripedpajamas:
@JL2210 @Xandernardo @Thilas I set up 32-bit Windows builds but can't test them myself. The build process went smoothly. I'll close this out and if anyone has any issues using them we can open up a new issue.

Thilas:
Thanks @stripedpajamas! I quicly tested it and it seems alright to me. VSCodium chocolatey package has also been updated in order to provide this 32-bit version as long with the existing 64-bit one.

PR #103, opened 2019-04-05
Build cleanup and Windows ia32

stripedpajamas:
Made `build.sh` easier to reason about and setup pipeline for building Windows 32-bit binaries. The binaries haven't been tested on 32-bit machines but the build process went smoothly.


stripedpajamas:
Closes #66

PR #124, opened 2019-04-14
Update logic to disable telemetry and crash reporting settings

### LINKED ISSUE #123: MS crash reporter still enabled on clean install
**Describe the bug**
Microsoft crash reporting appears to still be enabled upon clean installation

**To Reproduce**
1. Download and install VSCodium
2. Check telemetry settings in VSCodium UI

**Expected behavior**
Should be disabled out of the box. I am certain this was mentioned as being off in the docs but cannot find the exact quote at this time.

**Screenshots**
https://i.imgur.com/1cjhq6X.png

 - OS: Windows 10 pro
 - Architecture x64
 - Version 1.33.0+

**Additional context**

stripedpajamas:
Closes #123 

Default settings will be set properly on next release after merging this PR.

### DIFF
--- a/update_settings.sh
+++ b/update_settings.sh
@@ -1,9 +1,49 @@
-REPLACEMENT="s/'default': true/'default': false/"
+DEFAULT_TRUE="'default': true"
+DEFAULT_FALSE="'default': false"
+TELEMETRY_ENABLE="'telemetry.enableTelemetry':"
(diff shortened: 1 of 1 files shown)


stripedpajamas:
The crash reporting setting was moved to `src/vs/workbench/electron-browser/main.contribution.ts` which contained a lot of other settings, so the simplistic approach we had previously no longer worked. This new approach parses through a specified file for a "settings block" and then switches that setting to default to off instead of on.

Issue #139, opened 2019-04-28
Distribute versions for macOS as `dmg` files instead of zip


eirnym:
**Describe the bug**
Distribute versions for macOS as `dmg` files instead of zip. They have self-consistency check on mounting and more user-friendly features.

**Please confirm that this problem is VSCodium-specific**
- [X] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

**Please confirm that the issue/resolution isn't already documented**
- [X] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**To Reproduce**
1. Go to Downloads page
2. Check the latest distributions
4. See that version for macOS is distributed as zip archive.

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS
 - Architecture x64
 - Version 1.33.0

**Additional context**

stripedpajamas:
Hi @eirnym -- I like `dmg` better than `zip` as well. As Microsoft does not currently offer a `dmg` for Visual Studio Code, we would be starting from scratch. PRs are welcome toward this goal.

PR #162, opened 2019-05-17
Feature/DMG

RandyStevens:
This could be a potential solution to #139 

Uses a FOSS cli node app called create-dmg by @sindresorhus.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ Please note that some Visual Studio Code extensions have licenses that restrict 
 
 ## <a id="supported-os"></a>Supported OS
 - [x] OSX x64 (zipped app file)
+- [x] OSX x64 (DMG installer)
 - [x] Linux x64 (`.deb`, `.rpm`, `AppImage`, and `.tar.gz` files)
 - [x] Linux x86 (`.deb`, `.rpm`, and `.tar.gz` files)
 - [x] Windows x64
--- a/README.md
+++ b/README.md
@@ -92,8 +92,7 @@ For more information on getting all the telemetry disabled and tips for migratin
 Please note that some Visual Studio Code extensions have licenses that restrict their use to the official Visual Studio Code builds and therefore do not work with VSCodium. See [this note](https://github.com/VSCodium/vscodium/blob/master/DOCS.md#proprietary-debugging-tools) on the Docs page for what's been found so far and possible workarounds.
 
 ## <a id="supported-os"></a>Supported OS
-- [x] OSX x64 (zipped app file)
-- [x] OSX x64 (DMG installer)
+- [x] OSX x64 (zip, dmg)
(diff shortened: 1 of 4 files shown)


create_dmg.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+if [[ "$SHOULD_BUILD" == "yes" ]]; then
+  if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+    cd VSCode-darwin
+    create-dmg ../VSCodium-darwin-${LATEST_MS_TAG}.app
```
RandyStevens: I tried to follow your lead for this bashfile.

stripedpajamas: I think this should be `create-dmg VSCodium.app ..`

The syntax is `create-dmg <src> [dst]` and you want the generated `dmg` file as a sibling to the `zip` file

stripedpajamas:
This looks great @RandyStevens 🙌 🙌 🙌 

I tested it out locally and the generated `dmg` looks good. I think it will be a great addition to the available download options.

README.md
```diff
@@ -93,6 +93,7 @@ Please note that some Visual Studio Code extensions have licenses that restrict
 
 ## <a id="supported-os"></a>Supported OS
 - [x] OSX x64 (zipped app file)
+- [x] OSX x64 (DMG installer)
```
stripedpajamas: Combine these lines to be `OSX x64 (zip, dmg)`

create_dmg.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+if [[ "$SHOULD_BUILD" == "yes" ]]; then
+  if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+    cd VSCode-darwin
+    create-dmg VSCodium-darwin-${LATEST_MS_TAG}.app ../
```
RandyStevens: Here is the command with the correct syntax.

stripedpajamas: The app file here actually doesn't have the version, so just `VSCodium.app` (when we make the zip file is when the version gets added to the file name)

RandyStevens: Let me know if the version tag there should be dropped.

create_dmg.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+if [[ "$SHOULD_BUILD" == "yes" ]]; then
+  if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+    cd VSCode-darwin
+    create-dmg VSCodium.app ..
```
RandyStevens: Got it this time! 😅Thanks for the help!

stripedpajamas:
LGTM

PR #176, opened 2019-05-29
Launch the IDE with codium (instead of vscodium)

### LINKED ISSUE #36: codium as app executable name instead of vscodium?
The titles, names, and executables in the official and OSS versions all are variations of `code` and `code-oss` respectively. Maybe we should keep the pattern by making the executable `codium`? Also I think the  Visual Studio name is copyrighted by Microsoft, not sure, but we might as well be careful. Something else I like about the original `code` executable is that it's very quick to type, enabling commands like `code ~/very/big/path` feel more fluent and quick.

bittner:
We want to run VSCodium via `codium` in the terminal. This aligns with the non-free binary distribution and should make transitioning easier w.r.t. muscle memory.

Fixes #36.

### DIFF
--- a/VSCodium-AppImage-Recipe.yml
+++ b/VSCodium-AppImage-Recipe.yml
@@ -12,7 +12,7 @@ ingredients:
     - code
     - libgconf2-4
   dist: trusty
-  sources: 
+  sources:
     - deb http://archive.ubuntu.com/ubuntu/ trusty main universe
   script:
     - pwd
@@ -32,4 +32,4 @@ script:
   - convert vscodium.png -resize 32x32 usr/share/icons/hicolor/32x32/apps/vscodium.png
   - convert vscodium.png -resize 24x24 usr/share/icons/hicolor/24x24/apps/vscodium.png
   - convert vscodium.png -resize 22x22 usr/share/icons/hicolor/22x22/apps/vscodium.png
-  - ( cd usr/bin/ ; ln -s ../share/vscodium/vscodium  . )
+  - ( cd usr/bin/ ; ln -s ../share/vscodium/codium  . )
--- a/build.sh
+++ b/build.sh
@@ -21,7 +21,7 @@ if [[ "$SHOULD_BUILD" == "yes" ]]; then
   extensionsGallery='setpath(["extensionsGallery"]; {"serviceUrl": "https://marketplace.visualstudio.com/_apis/public/gallery", "cacheUrl": "https://vscode.blob.core.windows.net/gallery/index", "itemUrl": "https://marketplace.visualstudio.com/items"})'
   nameShort='setpath(["nameShort"]; "VSCodium")'
   nameLong='setpath(["nameLong"]; "VSCodium")'
-  applicationName='setpath(["applicationName"]; "vscodium")'
+  applicationName='setpath(["applicationName"]; "codium")'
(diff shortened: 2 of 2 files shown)


bittner:
I certainly also need to [adapt the symlink](https://github.com/VSCodium/vscodium/blob/master/VSCodium-AppImage-Recipe.yml#L35) in the AppImage recipe. I'm not sure if the changed `applicationName` value also affects some artifacts that are generated.

I'm a bit reluctant (read: lazy) to install the development dependencies on my machine and run the build locally (I know, shame on me! :grimacing:), so every helpful hint is appreciated.

bittner:
Can someone help me understand where Electron/Gulp builds the final binary? Looks like it's not built [inside the code tree](https://travis-ci.com/VSCodium/vscodium/jobs/203856439)... :worried: @stripedpajamas maybe?

I'm from a different planet, don't know the place, sorry! :grimacing:

stripedpajamas:
What OS are you building? It puts it in a different place for each OS. If you look at the deploy step in the Travis file you can see the paths we pull from (for instance, the Mac binary gets placed in a folder one level up from the vscode folder which we then zip).

stripedpajamas:
@bittner I recommend putting tests in a different PR to keep this PR scoped to issue #36 

Tests will be different based on the build OS.

bittner:
Sure. How do we verify that the changes actually do what we want?

(I'll remove the unneeded commits and squash everything to a single commit once we're done. No worries!)

stripedpajamas:
Unfortunately verifying the results takes a lot of hoop jumping. I am writing up a contributor's guide now, but it involves setting up a Travis environment with GH credentials and letting it build+deploy the binaries to your fork. 

Two options:
- wait a day or two for Contributor's guide and follow the instructions to test
- or I can merge this one because the change is small

If you have VSCodium installed, you should be able to edit your own product.json, change the application name to codium and then start it up. Then go to the command pallette and "install codium alias to the PATH" and then confirm it works in your terminal. More info on the latter part of those steps in the DOCS.md file in this repo.

Thanks for your help 👍

bittner:
I'd be fine if you merged the change right away. I've undone all commits except the ones doing the application name change.

I would guess the Debian package also has code that creates the symlink `/usr/bin/vscodium -> /usr/share/vscodium/bin/vscodium`. Where is this code? I didn't find it.

The "install codium alias to the PATH" (or similar) doesn't exist in my VSCodium command palette. Not sure why.

stripedpajamas:
Debian should be taken care of with the product.json change as well... the way MS has it set up, the product.json really is the source of truth. Debian post install where the symlink is created is [here](https://github.com/microsoft/vscode/blob/master/resources/linux/debian/postinst.template#L8), and [here](https://github.com/microsoft/vscode/blob/master/build/gulpfile.vscode.linux.js#L89) is where the build process replaces `@@NAME@@` with `product.applicationName`.

bittner:
Everyone installing VSCodium from packages, note:

* Now even the [package name changed](https://gitlab.com/paulcarroty/vscodium-deb-rpm-repo/commit/47585ae8decacc960f3a1f9a1e847b1893a26544) to `codium`.
* If you have a  `vscodium` package installed this will not be upgraded (it'll stay stuck at v1.34.0). You have to install the `codium` package explicitly:

```
sudo apt install codium
```
```
dnf install codium
```
```
zypper in codium
```

PR #179, opened 2019-05-30
change icon name to vscodium.png

### LINKED ISSUE #178: VSCodium icon name conflicts
In Linux, the icon name in the vscodium.desktop file is com.visualstudio.code.oss.png, which I believe is the same one that's used when building a freshly cloned vscode from source.
This makes icon themes overwrite vscodium's icon with vscode's one, as they cannot tell the difference between the two.
A simple fix would be to change the icon filename to something like vscodium.png

esclerofilo:
I believe this should fix #178. Unfortunately I haven't tested it because of a gulp error.


stripedpajamas:
@esclerofilo thanks for making this change. Looks good to me. Can you resolve the conflicts?

esclerofilo:
@stripedpajamas Ok, done

PR #241, opened 2019-08-15
Update DOCS.md

watkinsmatthewp:
Added details on how to edit the product.json to enable extensions


stripedpajamas:
Thanks @watkinsmatthewp 🙏 

Can you add an additional line in this block about how sometimes the licensing around these extensions are restrictive such that even altering the product.json can sometimes not help.

Issue #256, opened 2019-09-29
TravisCI switched build from Ubuntu 14 to Ubuntu 16 (builds incompatible with RHEL 7 and CentOS 7)


ajkerzner:
**Describe the bug**
"As of August, 13th 2019, we’ve switched the default Linux distribution on Travis CI from Ubuntu Trusty 14.04 LTS to Ubuntu Xenial 16.04. Here are the most common issues our customers ran into and how you can fix them." (Per https://docs.travis-ci.com/user/trusty-to-xenial-migration-guide )

Current builds of VS Codium are built with Ubuntu 16 (xenial) and gcc 5, and are incompatible with RHEL 7 and CentOS 7.

VS Code is built with Ubuntu 14 (trusty) and gcc 4, and _are_ compatible with RHEL 7 and CentOS 7.

**Please confirm that this problem is VSCodium-specific**
- [x] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

**Please confirm that the issue/resolution isn't already documented**
- [x] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**To Reproduce**
Run on RHEL 7 or CentOS 7, with `codium --verbose`.

**Expected behavior**
You will **not** see stuff like:
```
[main 2019-09-29T13:40:45.304Z] [uncaught exception in main]: Error: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /usr/share/codium/resources/app/node_modules.asar.unpacked/vscode-sqlite3/build/Release/sqlite.node)
[main 2019-09-29T13:40:45.304Z] Error: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /usr/share/codium/resources/app/node_modules.asar.unpacked/vscode-sqlite3/build/Release/sqlite.node)
[main 2019-09-29T13:40:50.297Z] { Error: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /usr/share/codium/resources/app/node_modules.asar.unpacked/vscode-sqlite3/build/Release/sqlite.node)
```
You will see a working Visual Studio Codium editor.

**Screenshots**

EXPECTED:
![vscodium-good-scrubbed](https://user-images.githubusercontent.com/11799692/65833501-216d4f00-e29f-11e9-8e8f-6769d5b58743.png)

ACTUAL:
[...]

PR #257, opened 2019-09-29
Update TravisCI config to use Ubuntu 14.04

ajkerzner:
"As of August, 13th 2019, we’ve switched the default Linux distribution on Travis CI from Ubuntu Trusty 14.04 LTS to Ubuntu Xenial 16.04. Here are the most common issues our customers ran into and how you can fix them." - https://docs.travis-ci.com/user/trusty-to-xenial-migration-guide

Tested and fixed the version of Ubuntu used by TravisCI to Ubuntu 14 (trusty), as the default changed in August (and broke compatibility with RHEL 7, CentOS 7, and deviated from VS Code's official builds).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ matrix:
     - os: linux
       sudo: required
       env: BUILDARCH=x64
+      dist: trusty
     - os: osx
 
 language: node_js


ajkerzner:
Fixes #256 .

PR #260, opened 2019-10-07
Solved punctuations and grammar

patelsneh18:
Loved the level of codes :-)


stripedpajamas:
I am sorry, @patelsneh18 -- I do not agree with the grammar decisions here and see no readability issues with how it currently is.

Issue #272, opened 2019-10-23
Building VSCodium Locally


kreck90:
Could you please help me to build it locally so I can improve it?

I'm trying to build it locally on Debian 10 (Buster) but for some reason it doesn't work.

I know I should use Ubuntu Trusty (as specified in YAML) but all other software like Node.js have supported version so this should not be an issue. Debian 10 is a clean install on a VM so there should be no conflicts.

I follow "instructions" in `.travis.yml` as non-root user that is member of sudo group like this:

``` sh
sudo apt-get -y install git jql nodejs npm

mkdir ~/build
cd ~/build

git clone https://github.com/VSCodium/vscodium.git

cd vscodium

curl -o- -L https://yarnpkg.com/install.sh | bash
export PATH="$HOME/.yarn/bin:$PATH"

# manually set some variables that are present in shell scripts
export TRAVIS_OS_NAME='linux'
export BUILDARCH='amd64'
export CI_WINDOWS='False'
export GITHUB_TOKEN='<my token>'

. install_deps.sh
. get_repo.sh
. check_tags.sh
```
I think there is something wrong when I run `check_tags.sh`  since I get:
~~~
...
    "created_at": "2019-10-16T10:32:35Z",
    "updated_at": "2019-10-16T10:32:36Z",
    "browser_download_url": "https://github.com/VSCodium/vscodium/releases/download/1.39.2/VSCodiumUserSetup-x64-1.39.2.exe.sha256"
  }
]
Already have all the Linux x64 builds
~~~
and when I run `echo "HAVE_64_RPM: $HAVE_64_RPM HAVE_64_DEB: $HAVE_64_DEB HAVE_64_TAR: $HAVE_64_TAR SHOULD_BUILD: $SHOULD_BUILD"` I get
~~~
HAVE_64_RPM: true HAVE_64_RPM: true HAVE_64_TAR: true SHOULD_BUILD:
~~~
* so in order to run following script I need to set the previous variables...
* this is what I do:
``` sh
export SHOULD_BUILD='yes'
export HAVE_64_RPM='false'
export HAVE_64_DEB='false'
export HAVE_64_TAR='false'

./build.sh
```
* but it doesn't work anyway, this is tail of the output
~~~
CSS INLINING IMAGE AT out-build/vs/workbench/contrib/comments/browser/media/reaction-dark.svg MORE THAN ONCE. CONSIDER CONSOLIDATING CSS RULES
[...]

PR #276, opened 2019-10-30
Create HOW_TO_BUILD_LOCALLY.md

kreck90:
Answer for https://github.com/VSCodium/vscodium/issues/272 question.

### DIFF
--- a/HOW_TO_BUILD_LOCALLY.md
+++ b/HOW_TO_BUILD_LOCALLY.md
@@ -1,0 +1,63 @@
+## Tested on Debian 10
+
+* Debian 10 has Node.js 10.15 and 10.16 is required
+  * install prerequisites for Node.js
+``` sh
+sudo apt-get update && sudo apt-get -y install python g++ gcc make clang wget
+```
+* build and install Node.js and npm
+```
+mkdir ~/build_node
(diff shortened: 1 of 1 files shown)


GitMensch:
Any update on this? moving it to a contrib folder sounds reasonable.

GitMensch:
@stripedpajamas as this is closed without a landing of the file anywhere but I'd like to have the info "somewhere" - What do you think of:

* adding a Wiki (that was possibly discussed before but I did not followed that so I don't know the state) where I can add a page for that
* adding that to the end of DOCS.md with a proper context
* create a new contrib folder, adding the initial file there, possibly reference in DOCS.md

daiyam:
I agree with @GitMensch. Wiki or new doc file are good options.

The `build/build_*` scripts are good start to manually build the project but we need to list the dependencies, add some disclaimers (8Go+ memory, can't be compiled under macOS 10.11, esbuild/go dependency).

Also, I would like to add a section how  to locally build the project in the same environment of Github Actions (can be useful when testing AppImage packaging...)

daiyam:
@GitMensch @stripedpajamas I've published the PR #754 to add some build doc

Issue #253, opened 2019-09-09
Linux AppImage Python extension path problem


adnion:
**Please confirm that this problem is VSCodium-specific**
- [x ] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.
It is APPIMAGE specific!

**Please confirm that the issue/resolution isn't already documented**
- [x ] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**To Reproduce**
1. Install Microsoft's python extension. 
2. Try to run python from vscodium's terminal. It uses the local python installation.
The error is "ModuleNotFoundError: No module named 'encodings'"
3. "printenv | grep PYTHON" gives PYTHONHOME and PYTHONPATH located in /tmp

**Expected behavior**
Python works.

**Temporary fix**
Remove AppRun and replace it with a symlink to usr/bin/codium

stripedpajamas:
@probonopd would you have any insight here? I'm not familiar with what kind of isolation AppImage might be doing.

probonopd:
`AppRun` exports a couple of paths, assuming that a private copy of Python is bundled inside the AppImage.

https://github.com/AppImage/AppImageKit/blob/fef038a6283701f6976589bf7becf975b020d453/src/AppRun.c#L169

https://github.com/AppImage/AppImageKit/blob/fef038a6283701f6976589bf7becf975b020d453/src/AppRun.c#L178

Do you want to use the Python that may or may not be on the system? Then do not use the `AppRun` binary but your own script or symlink.

tyu1996:
> AppRun exports a couple of paths, assuming that a private copy of Python is bundled inside the AppImage.

I ran by using the command inside terminal: 
`./VSCodium.AppImage .` 

And it opened VSCodium which inspects its own mounted directory.
![vscApp](https://user-images.githubusercontent.com/3956486/67560353-4c877900-f74d-11e9-9f7f-48cde8250f0a.png)
From the screenshot above, is it there is no python link in /usr/bin caused this issue? Both Python2 & Python3 doesn't work as mentioned above.

probonopd:
To the authors of VSCodium: Do you want to use the Python that may or may not be on the system?

tyu1996:
It's better to use the Python that exists in user's machine rather than packaging them inside the AppImage. 

[...]

PR #277, opened 2019-11-05
Update create_appimage.sh

tyu1996:
Addresses issue #253 
Modified pkg2appimage removes PYTHONPATH & PYTHONHOME exportations.

### DIFF
--- a/create_appimage.sh
+++ b/create_appimage.sh
@@ -3,8 +3,8 @@ if [[ "$BUILDARCH" == "x64" ]]; then
   # install a dep needed for this process
   sudo apt-get install desktop-file-utils
 
-  # download pkg2appimage from github
-  curl -LO "https://github.com/AppImage/pkg2appimage/raw/master/pkg2appimage"
+  # download modified pkg2appimage from gist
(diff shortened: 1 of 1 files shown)


create_appimage.sh
```diff
@@ -3,8 +3,8 @@ if [[ "$BUILDARCH" == "x64" ]]; then
   # install a dep needed for this process
   sudo apt-get install desktop-file-utils
 
-  # download pkg2appimage from github
-  curl -LO "https://github.com/AppImage/pkg2appimage/raw/master/pkg2appimage"
+  # download modified pkg2appimage from gist
+  curl -LO "https://gist.githubusercontent.com/tyu1996/39b8a7a29c2cdeef882daaa1702ec971/raw/2e788010deafb2f1214439ee0bc8e46685480e45/pkg2appimage"
```
stripedpajamas: Is there any way we can avoid doing this? What about if we put the modified pkg2appimage directly in this repo?

tyu1996: Is it something regarding the gist?

pkg2appimage requires functions.sh which requires AppRun to complete the AppImage packing, I modified all three files just for that. Is it encouraged if I put all these 3 files in this repo?

stripedpajamas: Yes; I would be ok keeping our modified pkg2appimage files in a folder in this repo.

tyu1996:
The latest PR continues this.

PR #278, opened 2019-11-06
Use cd ... || exit in case cd fails.

mztriz:


stripedpajamas:
Is this necessary? Have we seen builds where the `vscode` folder doesn't exist but we still got into the `build.sh` logic?

mztriz:
> Is this necessary? Have we seen builds where the `vscode` folder doesn't exist but we still got into the `build.sh` logic?

If/when cd fails, the script will keep going but do all its operations in the wrong directory, it could  be messy or lead to something unexpected.

PR #294, opened 2019-11-25
Modified AppImage build method

tyu1996:
Addresses the issue #253 , and cooperate with PR #277 , the modified pkg2appimage-set (included locally into the repo) fixes the PYTHONPATH problem.


stripedpajamas:
Thank you! We should see the fixed AppImage when the next release builds.

PR #298, opened 2019-12-05
Build natively on arm64

dimkr:


dimkr:
My mistake, intended to target `VSCodium:master` only once everything works

PR #302, opened 2019-12-09
Add hacky, initial arm64 support

dimkr:
The `build.sh` changes can be cleaned up a little bit thanks to https://github.com/microsoft/node-native-keymap/commit/e31be382147decfbdf943ba0e63acc920098fd06 and https://github.com/microsoft/vscode/pull/86659 (by using `PKG_CONFIG=pkg-config-aarch64-linux-gnu`), but it would be better to give something to those desiring a working arm64 build, then iterate from there.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,8 +2,8 @@
 
 function keep_alive() {
   while true; do
-    date
-    sleep 60
+    echo .
+    read -t 60 < /proc/self/fd/1 > /dev/null 2>&1
   done
 }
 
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -2,8 +2,8 @@
 
 function keep_alive() {
   while true; do
-    date
-    sleep 60
+    echo .
+    read -t 60 < /proc/self/fd/1 > /dev/null 2>&1
```
stripedpajamas: Why were these lines changed ? Can you help me understand what `read -t 60 < /proc/self/fd/1 > /dev/null 2>&1` does ?

stripedpajamas:
Looks great 👏 thank you so much for getting this together. Will merge, just had a quick question on one of the changes. Otherwise, after merging, we should see arm builds on next MS release.

build.sh
dimkr: When available RAM is low, the shell fails to run processes and exits. `date` and `sleep` are new processes. By using `echo` and `read` (from a pipe there's nothing to read from, the `stdout` pipe) with timeout, two shell built-in commands, we can output something and add delay, but without running external processes. This makes CI more reliable.

dimkr:
> after merging, we should see arm builds on next MS release.

You can manually trigger a build on `master` to add `arm64` packages to the current release :)

PR #305, opened 2019-12-12
Add an APT repository

dimkr:
https://packagecloud.io/dimkr/vscodium needs to be replaced with https://packagecloud.io/vscodium/vscodium (and its API token).

The free tier should be enough.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,8 +48,7 @@ deploy:
     package_glob: "*.deb"
     repository: vscodium
     username: dimkr
-    token:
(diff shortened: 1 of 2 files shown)


dimkr:
Seems the free plan is limited to 250 MB and not that much traffic. And the paid plans are expensive.

dimkr:
Trying https://github.com/dimkr/vscodium/commit/8b6540e65fa1e25d311fd0da525b679dcab39ca0, the free tier should be enough for one package per architecture :)

PR #328, opened 2020-02-14
product.json changes for Remote SSH

zhaofengli:
The two commits enable the `ms-vscode-remote.remote-ssh` API and set the `serverDataFolderName` key<sup>[1]</sup> in `product.json`, required for Microsoft's Remote SSH extension to function. Note that the server will try to validate the "client handshake," which you will have to patch manually at your own risk (mentioned in their [FAQ](https://github.com/microsoft/vscode/wiki/Differences-between-the-repository-and-Visual-Studio-Code#what-goes-where>)). This PR is what we can do from the editor side without violating any license.

Tested to work on x86-64, with Remote SSH connected to a x86-64 host.

<sup>[1]</sup> If `serverDataFolderName` is not set, the extension will always select the insider version of the server binary, causing a version mismatch when connecting.

### DIFF
--- a/prepare_vscode.sh
+++ b/prepare_vscode.sh
@@ -41,7 +41,7 @@ win32AppUserModelId='setpath(["win32AppUserModelId"]; "Microsoft.VSCodium")'
 win32ShellNameShort='setpath(["win32ShellNameShort"]; "VSCodium")'
 win32x64UserAppId='setpath (["win32x64UserAppId"]; "{{2E1F05D1-C245-4562-81EE-28188DB6FD17}")'
 urlProtocol='setpath(["urlProtocol"]; "vscodium")'
-extensionAllowedProposedApi='setpath(["extensionAllowedProposedApi"]; getpath(["extensionAllowedProposedApi"]) + ["ms-vsliveshare.vsliveshare"])'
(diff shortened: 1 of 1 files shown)


appelgriebsch:
Any updates on this? Would love to see this capability added to vscodium....

appelgriebsch:
After adding the entries to product.json I was able to initiate the SSH connection to a remote system, but the handshake with the vscode-server failed...

image

zhaofengli:
@appelgriebsch That's the second part (the "client handshake") that you will have to patch out from the server. Grep for `Unauthorized client refused` in `~/.vscode-server-oss` and you will see how (you might want to run the code through a prettifier first).

stripedpajamas:
Awesome, thanks for the explanation and the PR @zhaofengli 👍

AyyKamp:
> @appelgriebsch That's the second part (the "client handshake") that you will have to patch out from the server. Grep for `Unauthorized client refused` in `~/.vscode-server-oss` and you will see how (you might want to run the code through a prettifier first).

Could you explain exactly how to do this? I tried modifying remoteExtensionHostAgent.js`but it did not work :(

zhaofengli:
@AyyKamp You need to kill the server processes after changing the code, then reconnect from VSCodium.

PR #334, opened 2020-02-17
Added section on installing the snap

popey:
Thanks for making VS Codium. Now the codium snap is available in the stable channel of the snap store, I added a section to the README.md so people can find it. I also added a badge at the top which shows what build is current in the snap store, and the build status.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@
 [![build status](https://travis-ci.com/VSCodium/vscodium.svg?branch=master)](https://travis-ci.com/VSCodium/vscodium) 
 [![license](https://img.shields.io/github/license/VSCodium/vscodium.svg)](https://github.com/VSCodium/vscodium/blob/master/LICENSE)
 [![Gitter](https://img.shields.io/gitter/room/vscodium/vscodium.svg)](https://gitter.im/VSCodium/Lobby)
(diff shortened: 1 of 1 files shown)


stripedpajamas:
Has the snap been tested at all? The issue to make the snap is still open and I didn't create the codium one that's out there, so I want to get some clarity on what it is before adding instructions here.

popey:
I've tested it personally, and it's currently got nearly 250 users across a bunch of distros. 

Screenshot from 2020-02-27 11-30-16

Currently no issues filed about it on the [issue tracker](https://github.com/snapcrafters/codium/issues) for that package.

isgj:
I've been using VsCodium from snap from day one and not a single issue.

@popey maybe the command should be `sudo snap install --classic codium`

popey:
I'd recommend not specifying the `--classic` option. The user should be informed that it's a classic snap, which they will do if they don't copy/paste that in there too. As follows:

```
alan@mcp:~$ sudo snap install codium
error: This revision of snap "codium" was published using classic confinement
       and thus may perform arbitrary system changes outside of the security
       sandbox that snaps are usually confined to, which may put your system at
       risk.

       If you understand and want to proceed, repeat the command including
       --classic.
```

Also, numbers are still climbing nicely :)

Screenshot from 2020-03-21 17-18-43

PR #346, opened 2020-02-26
Add update information to AppImage

zilti:


stripedpajamas:
Can you explain what this change does? Not too familiar with how app images update. If we need the zsync file included in the GH release, the .travis.yml file would need an update too I think.

zilti:
Sure. Adding the `UPDATE_INFORMATION="gh-releases-zsync|VSCodium|vscodium|latest|*.AppImage.zsync"` environment variable causes the tool to embed that information into the AppImage. The format is `updatetype|organisation|repository|tag|zsyncfile`, and `latest` will cause it to use the newest tag/release. Tools like AppImageUpdate can then update the AppImage.

stripedpajamas:
Awesome, thanks so much for doing this

zilti:
I noticed that the .zsync file does not get added to the release... I am not sure why this happens though, I don't know your release pipeline very well

PR #354, opened 2020-03-12
refactor: extract prepare_vscode from build script

piedar:
I've been working on [an ebuild](https://github.com/piedar/overboard/tree/master/app-editors/vscodium) for vscodium so it can be installed in a Gentoo system and compiled from source (as much as possible).  Currently I have to patch your build script to remove steps which are unnecessary for my purposes.  This PR extracts the important steps to a new script `prepare_vscode.sh` so they can be run separately from the build.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -19,62 +19,14 @@ if [[ "$SHOULD_BUILD" == "yes" ]]; then
   echo "LATEST_MS_COMMIT: ${LATEST_MS_COMMIT}"
   echo "BUILD_SOURCEVERSION: ${BUILD_SOURCEVERSION}"
 
-  cp -rp src/* vscode/
-  cd vscode || exit
-
   export npm_config_arch="$BUILDARCH"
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -19,62 +19,14 @@ if [[ "$SHOULD_BUILD" == "yes" ]]; then
   echo "LATEST_MS_COMMIT: ${LATEST_MS_COMMIT}"
   echo "BUILD_SOURCEVERSION: ${BUILD_SOURCEVERSION}"
 
-  cp -rp src/* vscode/
-  cd vscode || exit
-
   export npm_config_arch="$BUILDARCH"
   export npm_config_target_arch="$BUILDARCH"
```
piedar: Maybe these exports only apply to the `yarn` step and could be removed, but I left them here also just in case.

stripedpajamas:
Thanks for doing this! I was starting to hate how bulky that build.sh file was getting.

PR #363, opened 2020-03-16
updates README instructions to disable brew analytics

sneak:
This updates the README to include instructions to disable homebrew's analytics before installing vscodium.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@
 #### <a id="install-with-brew"></a>Install with Brew (Mac)
 If you are on a Mac and have [Homebrew](https://brew.sh/) installed:
 ```bash
+brew analytics off
 brew cask install vscodium
 ```
 


stripedpajamas:
@sneak I don't think this makes sense to have within this repository. There are dozens of other information leaking areas in how VSCodium is downloaded (GH releases), how VSCodium is installed (brew, choco, scoop, Gitlab and the deb/rpm repo, etc), and within VSCodium itself (extension marketplace, ["online services"](https://code.visualstudio.com/docs/getstarted/telemetry#_managing-online-services), auto updates (as you pointed out in #364)).

This project isn't aiming to discourage all forms of information leakage across all of the different touch points involved. We are aiming to build Microsoft's vscode repository into freely-licensed binaries with a community-driven default config. The fact that VSCodium comes with Microsoft's telemetry disabled is a by-product of this goal, not representative of the goal itself.

sneak:
What's the harm in including this change?  Homebrew unethically assumes telemetry consent; a great many people have no idea that Homebrew is spying for Google on each and every one of their downloads.

The other download methods you mention are not used for telemetry/tracking purposes.  Homebrew explicitly calls out to Google Analytics with a unique installation ID whenever a package is installed.  If a user is using VSCodium instead of VS Code, it can be reasonably assumed they do not wish to have telemetry enabled by default.  Homebrew's telemetry is enabled by default, so your *recommended method* of installing VSCodium *involves the sending of telemetry*.

At that point, why bother disabling telemetry in VS Code?

stripedpajamas:
This project doesn't have an opinion on the data Homebrew sends by default. As such we will not be suggesting that anyone [permanently opt-out of Homebrew's analytics](https://docs.brew.sh/Analytics#opting-out). Homebrew's analytics aren't a secret, and it's not in any way this project's responsibility to broadcast them.

PR #404, opened 2020-05-18
Use open-vsx.org as extension gallery

### LINKED ISSUE #372: Integrate with Open VSX Registry
Please integrate with Open VSX Registry, open-vsx.org. It provides extensions for VS Code compatible editors in a community driven way.

Example from Theia:
https://github.com/eclipse-theia/theia/commit/73bd544d99a1f93159cc83cedd9769e627033811

spoenemann:
Fixes #372.

The Microsoft Marketplace [Terms of Use](https://aka.ms/vsmarketplace-ToU) state

> Marketplace Offerings are intended for use only with Visual Studio Products and Services and you may only install and use Marketplace Offerings with Visual Studio Products and Services.

AFAICT VSCodium is not a "Visual Studio Product" and thus it may not use any "Marketplace Offerings". Strangely, the question whether the extensionsGallery setting can be used in forks (https://github.com/Microsoft/vscode/issues/31168) has been left unanswered for almost three years by the VS Code team. I wonder whether that's a deliberate decision to avoid annoying open source communities with legal issues. In any case, I think it's better to work around such issues and switch to an open source marketplace (see also https://github.com/git-for-windows/build-extra/pull/278#issuecomment-622045223).

Open VSX now has an adapter to the VS Code Marketplace API (https://github.com/eclipse/openvsx/issues/55). I also added the `linkProtectionTrustedDomains` property to allow forwarding to the extension detail page on open-vsx.org when the user clicks on the name of an extension.

See [this article](https://www.gitpod.io/blog/open-vsx/) for more information on the motivation behind Open VSX.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -32,7 +32,20 @@ _(For example the C# extension `ms-vscode.csharp` sends tracking data to Microso
 
 ## <a id="extensions-marketplace"></a>Extensions + Marketplace
 
-Until something more open comes around, we use the Microsoft Marketplace/Extensions in the `product.json` file. Those links are licensed under MIT as per [the comments on this issue.](https://github.com/Microsoft/vscode/issues/31168#issuecomment-317319063)
+The `product.json` file is set up to use [open-vsx.org](https://open-vsx.org/) as extension gallery, which has an [adapter](https://github.com/eclipse/openvsx/wiki/Using-Open-VSX-in-VS-Code) to the Marketplace API used by VS Code. Since that is a rather new project, you will likely miss some extensions you know from the VS Code Marketplace. You have the following options to obtain such missing extensions:
+
+ * Ask the extension maintainers to publish to [open-vsx.org](https://open-vsx.org/) in addition to the VS Code Marketplace. The publishing process is documented in the [Open VSX Wiki](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions).
+ * Create a pull request to [this repository](https://github.com/open-vsx/publish-extensions) to have the [@open-vsx](https://github.com/open-vsx) service account publish the extensions for you.
+ * Download and [install the vsix files](https://code.visualstudio.com/docs/editor/extension-gallery#_install-from-a-vsix).
+ * Modify the `extensionsGallery` section of the `product.json` file in your VSCodium installation to use the VS Code Marketplace as shown below. However, note that [it is not clear whether this is legal](https://github.com/microsoft/vscode/issues/31168).
+```
+"extensionsGallery": {
+    "serviceUrl": "https://marketplace.visualstudio.com/_apis/public/gallery",
+    "itemUrl": "https://marketplace.visualstudio.com/items"
+}
+```
+
+See [this article](https://www.gitpod.io/blog/open-vsx/) for more information on the motivation behind Open VSX.
 
 ### Proprietary Debugging Tools
 
--- a/README.md
+++ b/README.md
@@ -106,7 +106,9 @@ Microsoft's build process (which we are running to build the binaries) does down
 ### Documentation
 For more information on getting all the telemetry disabled and tips for migrating from Visual Studio Code to VSCodium, have a look at this [Docs](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) page.
 
-### Extension Licensing
+### Extensions and the Marketplace
+According to the VS Code Marketplace [Terms of Use](https://aka.ms/vsmarketplace-ToU), _you may only install and use Marketplace Offerings with Visual Studio Products and Services._ For this reason, VSCodium uses [open-vsx.org](https://open-vsx.org/), an open source registry for VS Code extensions. See the [Extensions + Marketplace](https://github.com/VSCodium/vscodium/blob/master/DOCS.md#extensions-marketplace) section on the Docs page for more details.
+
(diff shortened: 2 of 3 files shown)


stripedpajamas:
Hi @spoenemann 👋 

I'm so excited about this, thank you so much for opening this PR!!

Can you help me understand a bit about how this works? I updated my product.json manually to use open-vsx to try it, and so far the experience is mostly seamless. I searched for `javascript` and found JavaScript extensions. I searched for `ava` and found no extensions, but I know there is an `ava` extension in the normal marketplace. Should users expect all VSCode Marketplace extensions to be present in open-vsx? If not, can you update the PR to make this clear in the README, and perhaps link to our DOCS.md for information about how to re-enable the default marketplace? I want to avoid a regression in user experience, especially given that a lot of users will receive this change automatically when their installation autoupdates (after we merge).

In fact, including any extra links about motivation / how it works etc right in the README (like the links you have in the PR summary) would be great as well. I'm guessing because open-vsx is so new, it will be unfamiliar for most of the userbase.

spoenemann:
Hi @stripedpajamas! Open VSX is completely separate from the VS Code Marketplace, so you cannot expect all Marketplace extensions to be present in Open VSX. In fact there are currently only 218 extensions because the service is quite new and still in the process of being transferred to the [Eclipse Foundation](https://www.eclipse.org) (the source code is already an Eclipse Foundation project, but the website is currently operated by [TypeFox](https://www.typefox.io)).

The idea is to build a new community of extension developers and consumers and convince maintainers of popular extensions to publish to Open VSX in addition to the VS Code Marketplace. If someone is missing a particular extension, they should first approach the maintainers of that extension and ask to publish it to Open VSX. In case the maintainers won't do that, the fallback solution is to add an entry to `extensions.json` in [open-vsx/publish-extensions](https://github.com/open-vsx/publish-extensions). A GitHub action will then take care of publishing the extension.

A regression in user experience is inevitable. Building a new community takes time. I agree more documentation on this will be helpful.

spoenemann:
I updated README.md and DOCS.md, WDYT?

DOCS.md
```diff
@@ -32,7 +32,20 @@ _(For example the C# extension `ms-vscode.csharp` sends tracking data to Microso
...
+The `product.json` file is set up to use [open-vsx.org](https://open-vsx.org/) as extension gallery, which has an [adapter](https://github.com/eclipse/openvsx/wiki/Using-Open-VSX-in-VS-Code) to the Marketplace API used by VS Code. Since that is a rather new project, you will likely miss some extensions you know from the VS Code Marketplace. You have the following options to obtain such missing extensions:
+
+ * Ask the extension maintainers to publish to [open-vsx.org](https://open-vsx.org/) in addition to the VS Code Marketplace. The publishing process is documented in the [Open VSX Wiki](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions).
+ * Create a pull request to [this repository](https://github.com/open-vsx/publish-extensions) to have the [@open-vsx](https://github.com/open-vsx) service account publish the extensions for you.
+ * Download and [install the vsix files](https://code.visualstudio.com/docs/editor/extension-gallery#_install-from-a-vsix).
+ * Modify the `extensionsGallery` section of the `product.json` file in your VSCodium installation to use the VS Code Marketplace as shown below.
```
akosyakov, line 40: Maybe good idea to mention https://github.com/Microsoft/vscode/issues/31168#issuecomment-317319063 here? It still looks questionable whether it is legal.

stripedpajamas:
Thanks for the clarification and the extra verbiage! Looks good to me. I'm super pumped about this initiative 😍

GitMensch:
@stripedpajamas  That is such an important change, can we get a new VSCodium release sooner than the next VSCode update?

stripedpajamas:
@GitMensch because of how things are setup currently, cutting a new VSCodium release would mean deleting the assets from the current release and triggering a new build. There are a bunch of downstream systems that will not react will well to that due to the new hashes. We don't currently have a way of releasing a version between VSCode versions.

GitMensch:
Thank you for the note, looks like this may be a good issue to track for "some later date" (seems useful to be independent from upcoming vscode releases, if necessary).

spoenemann:
If you'd like to chat more about this, feel free to share your thoughts at https://gitter.im/eclipse/openvsx

den32ds:
@spoenemann Nice request, but what should I do with Russian language support? What to do with the support of programming languages?

GitMensch:
I've actually stumbled over the language stuff, too - I think it is best to ask Open VSX to auto-import it as it was done for other stuff .... if those extensions have a free license, which I did not recheck.
@den32ds For the support of programming languages: ask the maintainers of the extensions to co-publish on the open vsx registry, many will do this quite fast. Note: some plugins - especially the ones from Microsoft do not have a license that allows you to use it anywhere else then the "VSCode" binaries created by Microsoft.

den32ds:
@GitMensch Do you suggest that I come up to everyone and say "please build for OpenVSX too"? ok, but why doesn’t it work from VSIX?

GitMensch:
> Do you suggest that I come up to everyone and say "please build for OpenVSX too"?

No, I've just set if you want to see it in the open vsx registry which is also what is searched in the vscodium extensions tab then it is likely best to ask the maintainer of the extension (or stand still and wait until others do).
You can of course manually download and install the extensions - most will work.

> but why doesn’t it work from VSIX?

I don't know what doesn't work for you but this definitely is quite a topic drift to this issue - you may want to create a new one after checking what the [README](https://github.com/VSCodium/vscodium#extensions-and-the-marketplace) and linked documentation says on this.

den32ds:
> I don't know what doesn't work for you but this definitely is quite a topic drift to this issue - you may want to create a new one after checking what the README and linked documentation says on this.

oh, thanks

PR #405, opened 2020-05-20
Add update information for appimage

linsui:
https://github.com/VSCodium/vscodium/pull/346 use a env variable `UPDATE_INFORMATION` but I can't find docs about it and it doesn't work. So I follow https://docs.appimage.org/packaging-guide/optional/updates.html#using-appimagetool instead. It works locally.

```
+ mv VSCodium-1.45.0-1588934730.glibc2.16-x86_64.AppImage VSCodium-1.45.0-1588934730.glibc2.16-x86_64.AppImage.zsync ../out/
+ ls -lh ../out/VSCodium-1.45.0-1588934730.glibc2.16-x86_64.AppImage ../out/VSCodium-1.45.0-1588934730.glibc2.16-x86_64.AppImage.zsync
-rwxr-xr-x 1 ult ult  89M 5月  20 17:07 ../out/VSCodium-1.45.0-1588934730.glibc2.16-x86_64.AppImage
-rw-rw-r-- 1 ult ult 311K 5月  20 17:07 ../out/VSCodium-1.45.0-1588934730.glibc2.16-x86_64.AppImage.zsync
```

### DIFF
--- a/src/resources/linux/appimage/functions.sh
+++ b/src/resources/linux/appimage/functions.sh
@@ -241,9 +241,9 @@ generate_type2_appimage()
     tar xf data.tar.gz
     sudo chown -R $USER .gnu*
     mv $HOME/.gnu* $HOME/.gnu_old ; mv .gnu* $HOME/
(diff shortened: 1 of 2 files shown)


stripedpajamas:
Looks great, and good catch with that `]` bug!

PR #431, opened 2020-06-19
Fix sqlite error on ARM build

tuanpham-dev:
Fix vscode sqlite error that makes welcome messages (notification, extension welcome pages) appear every time on vscode startup.

### DIFF
--- a/prepare_vscode.sh
+++ b/prepare_vscode.sh
@@ -13,6 +13,14 @@ patch -u src/vs/platform/update/electron-main/updateService.win32.ts -i ../patch
 
 yarn --frozen-lockfile
 yarn postinstall
+
+if [[ "$BUILDARCH" == *"arm"* ]]; then
+  sed -i -z 's/,\n[^\n]*arm[^\n]*//' node_modules/vscode-sqlite3/binding.gyp
+  sed -i "s/Release\/sqlite'/Release\/sqlite.node'/" node_modules/vscode-sqlite3/lib/sqlite3.js
+  yarn add -D electron-rebuild
+  npx electron-rebuild -f -w vscode-sqlite3
+fi
+
 mv product.json product.json.bak
 
 # set fields in product.json


prepare_vscode.sh
```diff
@@ -13,6 +13,14 @@ patch -u src/vs/platform/update/electron-main/updateService.win32.ts -i ../patch
 
 yarn --frozen-lockfile
 yarn postinstall
+
+if [[ "$BUILDARCH" == *"arm"* ]]; then
+  sed -i -z 's/,\n[^\n]*arm[^\n]*//' node_modules/vscode-sqlite3/binding.gyp
+  sed -i "s/Release\/sqlite'/Release\/sqlite.node'/" node_modules/vscode-sqlite3/lib/sqlite3.js
+  yarn add -D electron-rebuild
+  npx electron-rebuild -f -w vscode-sqlite3
```
stripedpajamas, line 21: if using `npx` there should be no reason to save `electron-rebuild` as a devDep on L20

tuanpham-dev: Do you mean we can omit this line?  `yarn add -D electron-rebuild`

stripedpajamas: yes

tuanpham-dev: I don't think removing that line works. I got this error when trying to use `npx` only
```
Step 29/36 : RUN npx electron-rebuild -f -w vscode-sqlite3
 ---> Running in d8195f16eb1b
npx: installed 164 in 38.491s

An unhandled error occurred inside electron-rebuild
Unable to find electron's version number, either install it or specify an explicit version

Error: Unable to find electron's version number, either install it or specify an explicit version
    at /root/.npm/_npx/1/lib/node_modules/electron-rebuild/lib/src/cli.js:87:19
    at Generator.next (<anonymous>)
    at fulfilled (/root/.npm/_npx/1/lib/node_modules/electron-rebuild/lib/src/cli.js:6:58)
```

stripedpajamas: I see... I guess electron-rebuild does some kind of install-time logic. LGTM

PR #432, opened 2020-06-20
Add instruction to install with Windows Package Manager (WinGet)

### LINKED ISSUE #425: Add winget installation method to documentation
I took the liberty of placing a VSCodium manifest in https://github.com/microsoft/winget-pkgs and thought that it could be worth mentioning here officially.

The problem with winget is that the process of submitting a PR to the repo is quite slow, so there can be up to a week of delay between an official release here, and it becoming available in winget. If that is not desirable then it could be discussed here.

cstayyab:
As VSCodium is available to install using Windows Package Manager, it's instructions to install should also be added to README file and that's what I did. I hope this can be merged without any problem.

Fixed #425

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@
 ## Table of Contents
 - [Download/Install](#download-install)
   - [Install with Brew](#install-with-brew)
+  - [Install with Windows Package Manager (WinGet)](#install-with-winget)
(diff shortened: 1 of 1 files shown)


stripedpajamas:
Thanks @cstayyab !!

Issue #471, opened 2020-08-15
Too big icon on start menu


linsui:
**Describe the bug**
Too big icon on start menu.

**Please confirm that this problem is VSCodium-specific**
- [x] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

**Please confirm that the issue/resolution isn't already documented**
- [x] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**Screenshots**
![sshot-009](https://user-images.githubusercontent.com/36977733/90304210-cdfaf400-dea4-11ea-8d2b-021caa639412.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Architecture x64
 - Version 1.48.0

Bufkin:
Same issue on Windows 10 Pro

linsui:
https://github.com/VSCodium/vscodium/blob/master/src/resources/win32/code_150x150.png
https://github.com/VSCodium/vscodium/blob/master/src/resources/win32/code_70x70.png
These icons should be fixed.

layday:
The new icon is oversized on Mac as well.

estatra:
Thanks for the reports. [Here's the fix for Windows](https://github.com/VSCodium/vscodium/pull/474).

For the 70x70 one I couldn't do a perfect 66% padding because the image editor doesn't scale to floating points, so I got it to as close as possible at 65.719% padding.

---
@layday Could specify what icon(s) please? Dock? Tab? Other?

@linsui @stripedpajamas Can you guys help me locate the problematic icons for Mac? I use Linux only. I recommend we solve this before merging, so I can append the Mac fix to the PR.

layday:
It appears to be the case that it is slightly oversized in the dock as well as in the Finder and the tab switcher in comparison to other icons.

<img width="418" alt="Screenshot 2020-08-16 at 00 15 49" src="https://user-images.githubusercontent.com/31134424/90321792-bdef1d00-df55-11ea-8147-d07942676725.png">

estatra:
Thanks for the quick reply layday. Seems to be a padding issue also. I'll wait for linsui and stripedpajamas help.

linsui:
I don't use mac. The icons looks perfectly on Windows. Thanks! @estatra

estatra:
> I don't use mac. The icons looks perfectly on Windows. Thanks! @estatra

No problem!

[...]

PR #474, opened 2020-08-15
Fix icon padding for Windows icons

estatra:


stripedpajamas:
thanks @estatra 👍

estatra:
> thanks @estatra

No problem! Please check https://github.com/VSCodium/vscodium/issues/471 when you have the time. I think I need to fix Mac and Linux ones as well, but need some help/confirmation.

PR #525, opened 2020-10-06
Fix arm builds

### LINKED ISSUE #520: GitHub Actions ARM Builds Fail
**Describe the bug**
Under releases there is none for arm with the expected name of `VSCodium-linux-arm-1.49.2.tar.gz`

stripedpajamas:
Turns out MS is maintaining docker images for Linux build environments, so have [forked those](https://github.com/VSCodium/vscode-linux-build-agent) and made them public: https://hub.docker.com/repository/docker/vscodium/vscodium-linux-build-agent

RPM for arm is busted until [this commit upstream](https://github.com/microsoft/vscode/commit/034f6fc5ba9f0f00983f2aeb559af411b2fb22e5) is released. Included in the next release will most likely also be [this commit](https://github.com/microsoft/vscode/commit/47d8a2560605b5b6b013719f7f385ec9750cae9f) which renames the gulp task for armhf builds; so I expect the need for a patch on next release to make armhf build here.

Should fix #520

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        vscode_arch: [x64, arm64, armhf]
+        vscode_arch: [x64, arm64, arm]
         include:
         - vscode_arch: x64
           npm_arch: x64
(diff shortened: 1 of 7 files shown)


stripedpajamas:
the arm check failures are somewhat expected; the build didn't find ARM artifacts in the repo and so built them. it failed to release them because this is a PR and the build doesn't have permission to release.

TODO: make release conditional on being on the main repo (vscodium/vscodium) so that PRs can build without error.

PR #531, opened 2020-10-09
Fix armhf and x64

stripedpajamas:
x64 builds were broken due to missing dependency (`convert`); I added that to the build environment (https://github.com/VSCodium/vscode-linux-build-agent/commit/e146653aa186ff2bc2a2dae7c5297beb6c11e802), but AppImage creation is broken for Docker-specific reasons. I will open an issue to track that and get some help from people who know more about AppImages.

For now this should get the deb/rpms for x64 and armhf for latest release back to working state.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


stripedpajamas:
Tracking AppImage issue in https://github.com/VSCodium/vscodium/issues/532

PR #537, opened 2020-10-19
Custom extensions gallery

### LINKED ISSUE #519: Switching between marketplaces
Hello,  
I am a user of vscodium since about two years, I really like this fork of vscode, but lately I have some issues with it.   
Most of them are caused by switching from proprietary store to [open-vsx](https://open-vsx.org/). I don't criticize this decision, but I want you to know how many troubles and annoyances it causes for regular users (I know using Microsoft's store can be problematic from legal point of view).

As a new user I would have no clue, why there is no access to some extensions. 
After looking for solutions in many issues and discussions among repository (it seems to be a time-consuming alternative for just downloading regular vscode) it appears that the simplest way of getting them is modification of file `product.json` https://github.com/VSCodium/vscodium/issues/418.  
It would be just one time inconvenience for loyal vscodium users, but it could be barrier for newcomers wanting to leave Microsoft's version.

This is problem for many users, the best measurement is number of issues describing it: https://github.com/VSCodium/vscodium/issues/514 https://github.com/VSCodium/vscodium/issues/508 https://github.com/VSCodium/vscodium/issues/493 https://github.com/VSCodium/vscodium/issues/483  https://github.com/VSCodium/vscodium/issues/463 etc...

Unluckily it appears that after every update (in my case I use PPA for regular Ubuntu package) Marketplace is broken and after every update I have to mess with `product.json` and this becomes a major inconvenience.

In this case I would like to suggest some (at least partial) solutions for this problem:
- Maybe it would be good to pin an issue/add note in README with instructions how to edit this file
- ensure preservation of applied changes in `product.json` between updates (seems to be not very invasive)
- add setting for using secondary marketplace (It seems to be invasive change into codebase, but it could be worth it in case of creation of another open-source marketplaces).

articlecat:
This pull request lets VSCodium accept 4 optional environment variables:
* `VSCODIUM_EXTENSIONS_GALLERY_SERVICE_URL`
* `VSCODIUM_EXTENSIONS_GALLERY_ITEM_URL`
* `VSCODIUM_EXTENSIONS_GALLERY_CONTROL_URL`
* `VSCODIUM_EXTENSIONS_GALLERY_RECOMMENDATIONS_URL`

If defined, these environment variables allow VSCodium to use a marketplace other than the default marketplace specified in [prepare_vscode.sh](https://github.com/VSCodium/vscodium/blob/master/prepare_vscode.sh) (currently [Open VSX](https://github.com/eclipse/openvsx)). In particular, they enable users to select the [Visual Studio Marketplace](https://marketplace.visualstudio.com/) with the `serviceUrl` and `itemUrl` values provided in microsoft/vscode#31168 (published under the MIT License in microsoft/vscode@b00945fc8c79f6db74b280ef53eba060ed9a1388).

I was able to build VSCodium with the changes in this pull request using the discussion in #272 as a guide, and I can confirm that the built VSCodium AppImage is able to access the Visual Studio Marketplace with `VSCODIUM_EXTENSIONS_GALLERY_SERVICE_URL` and `VSCODIUM_EXTENSIONS_GALLERY_ITEM_URL` set to the correct values.

This pull request is based on the interfaces defined in [productService.ts](https://github.com/microsoft/vscode/blob/master/src/vs/platform/product/common/productService.ts), and is inspired by flathub/com.visualstudio.code-oss@f4fb6f33f410a59a04a5b564792e024b524fd2cc (explained in flathub/com.visualstudio.code-oss#11) from the ["Code - OSS" Flatpak](https://github.com/flathub/com.visualstudio.code-oss).

Resolves #519.

### DIFF
--- a/patches/custom-extensions-gallery.patch
+++ b/patches/custom-extensions-gallery.patch
@@ -1,0 +1,28 @@
+diff --git a/src/vs/platform/product/common/product.ts b/src/vs/platform/product/common/product.ts
+index 2bea85740c..d0e9da036c 100644
+--- a/src/vs/platform/product/common/product.ts
++++ b/src/vs/platform/product/common/product.ts
+@@ -57,6 +57,23 @@ else {
+ 		});
+ 	}
+ 
++	// Set user-defined extension gallery
++	if (
++		env['VSCODIUM_EXTENSIONS_GALLERY_SERVICE_URL']
++		|| env['VSCODIUM_EXTENSIONS_GALLERY_ITEM_URL']
++		|| env['VSCODIUM_EXTENSIONS_GALLERY_CONTROL_URL']
++		|| env['VSCODIUM_EXTENSIONS_GALLERY_RECOMMENDATIONS_URL']
++	) {
++		Object.assign(product, {
++			extensionsGallery: {
++				serviceUrl: env['VSCODIUM_EXTENSIONS_GALLERY_SERVICE_URL'],
++				itemUrl: env['VSCODIUM_EXTENSIONS_GALLERY_ITEM_URL'],
++				controlUrl: env['VSCODIUM_EXTENSIONS_GALLERY_CONTROL_URL'],
++				recommendationsUrl: env['VSCODIUM_EXTENSIONS_GALLERY_RECOMMENDATIONS_URL']
++			}
++		})
++	}
++
+ 	Object.assign(product, {
+ 		version: pkg.version
+ 	});
--- a/prepare_vscode.sh
+++ b/prepare_vscode.sh
@@ -13,6 +13,7 @@ cd vscode || exit
 ../update_settings.sh
 
 # apply patches
+patch -u src/vs/platform/product/common/product.ts -i ../patches/custom-extensions-gallery.patch
 patch -u src/vs/platform/update/electron-main/updateService.win32.ts -i ../patches/update-cache-path.patch
 
 if [[ "$OS_NAME" == "osx" ]]; then


patches/custom-extensions-gallery.patch
```diff
@@ -0,0 +1,28 @@
...
+ 
++	// Set user-defined extension gallery
++	if (
++		env['VSCODIUM_EXTENSIONS_GALLERY_SERVICE_URL']
++		|| env['VSCODIUM_EXTENSIONS_GALLERY_ITEM_URL']
++		|| env['VSCODIUM_EXTENSIONS_GALLERY_CONTROL_URL']
++		|| env['VSCODIUM_EXTENSIONS_GALLERY_RECOMMENDATIONS_URL']
++	) {
++		Object.assign(product, {
++			extensionsGallery: {
```
gasinvein, line 17: Using `||` here doesn't seem safe, since some of the vars are mandatory. Maybe check if `_SERVICE_URL` and `_ITEM_URL` are both set, and proceed with overriding `extensionsGallery` only if they are?

pktiuk:
I think this PR should also describe this option in: [DOCS.md](https://github.com/VSCodium/vscodium/blob/master/DOCS.md#extensions-marketplace)

stripedpajamas:
This is an awesome idea. As @pktiuk mentioned, it would be great if PR included an updated DOCS.md file. I expect we would want to provide users a straightforward way to set their marketplace to the MS urls, maybe a shell code block example they can paste into their bash_profile/etc.

Additionally, as @GitMensch mentioned... I'm wondering if it would be helpful (for people who want to piecemeal replace) to allow per-property overrides. Something like:

```javascript
  // Allow user-defined extension gallery URLs
  const { serviceUrl, itemUrl, controlUrl, reccomendationsUrl } = product.extensionsGallery || {}
  Object.assign(product, {
    extensionsGallery: {
      serviceUrl: env['VSCODIUM_EXTENSIONS_GALLERY_SERVICE_URL'] || serviceUrl,
      itemUrl: env['VSCODIUM_EXTENSIONS_GALLERY_ITEM_URL'] || itemUrl,
      controlUrl: env['VSCODIUM_EXTENSIONS_GALLERY_CONTROL_URL'] || controlUrl,
      recommendationsUrl: env['VSCODIUM_EXTENSIONS_GALLERY_RECOMMENDATIONS_URL'] || recommendationsUrl
    }
  })
```

GitMensch:
... and please use `VSCODE_EXTENSIONS_GALLERY_SERVICE_URL` and friends.

GitMensch:
Any update on this? It did not got into 1.50.0, but maybe we can get it into the next iteration.

misaka00251:
Any update on this?

flux242:
please do SOMETHING with this one! I'm using an Appimage and can't change the product.json manually! There are several solutions proposed already, just decide already. I would switch immediately back to vscode to get extensions I need if they would provide an Appimage but they don't

flux242:
please note that this trick with the environment variables won't work with the Appimage when Appimage is used with firejail. To my opinion solution where the user defined version of product.json is checked first would work for all scenarios.

GitMensch:
> To my opinion solution where the user defined version of product.json is checked first would work for all scenarios.

And _how_ should this be done? product.json is read from its default place, so we'd need an option to say "take it from somewhere else" and this would be done exactly how without an environment variable? The only thing I can think of is an extra command line option like `--product=/path/to/product.json`, which may or may not work (depending where it is read, I assume it is read very early; and if we can "update" the data read from there after the initial load).
... in any case a new command line option would be much more work than changing specific values (or even the path to product.json) by an environment value,

flux242:
huh? like this
```
if (exist(~/.config/VSCodium/product.json)) {
   open('~/.config/VSCodium/product.json');
}
else {
  open("DEFAULT_LOCATION/product.json)
}
```

GitMensch:
@flux242 that's a completely different approach, that's also reasonable to merge. Can you please create a new PR with those changes?

GitMensch:
@articlecat Do you plan on working on this PR or should I pull that locally and create a new one (with `VSCODE` prefix, partial replacement and DOCS.md)?

coxackie:
@GitMensch please go ahead with a new PR, for the sake of all of us. One month is enough of a wait. @articlecat is AWOL, so you should take over.

daiyam:
I've done the PR #674 which allows the environment variables and custom `product.json`.

GitMensch:
Closed in favor of #674 - thanks @daiyam for finishing this nearly completed PR.

articlecat:
Thank you for completing this, @daiyam! I'm sorry for not getting around to this, but I'm happy to see this finished.

Issue #533, opened 2020-10-11
Set update.mode = none on install


ggreif:
**Describe the bug**

`codium` sometimes starts to download an update from a Microsoft server.
The corresponding setting ID is `update.mode` the value it is set originally is
``` js
"update.mode": "default"
```

**Please confirm that this problem is VSCodium-specific**
- [ ] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

NACK. We probably want to deviate from VSC here.

**Please confirm that the issue/resolution isn't already documented**
- [ ] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

https://github.com/VSCodium/vscodium/blob/master/DOCS.md#getting-all-the-telemetry-out mentioned that these settings can be disabled. I'm just asking whether they could come out of the box disabled, like this:
``` js
"update.mode": "none"
```

**To Reproduce**
1. Start `codium` the first time
2. Observe the bottom-left cogwheel, notifying you
3. Hover over it
4. See "Manage - New VSCodium update available" notice:
<img width="171" alt="Screen Shot 2020-10-11 at 12 40 21" src="https://user-images.githubusercontent.com/1312006/95676621-b3e94380-0bbf-11eb-887b-3855422bf949.png">
5. When clicking on it I get a menu offering update:
<img width="210" alt="Screen Shot 2020-10-11 at 12 41 40" src="https://user-images.githubusercontent.com/1312006/95676617-ad5acc00-0bbf-11eb-85c4-2fc8c79337d3.png">

**Expected behavior**

Come to life like this:

<img width="473" alt="Screen Shot 2020-10-11 at 12 47 09" src="https://user-images.githubusercontent.com/1312006/95676663-f9a60c00-0bbf-11eb-9959-869c8eda07e9.png">

**Screenshots**
See above.

**Desktop (please complete the following information):**
 - OS: Mac OS (High Sierra), Ubuntu 20.04
 - Architecture x64
 - Version 1.49.3

**Additional context**

For the record, `v1.50.0` just came out, so maybe that's the one it wants to download...

In both OSs I used `nixpkgs` to get the app.

ggreif:
For the record, `v1.50.0` just came out, so maybe that's the one it wants to download...

stripedpajamas:
Hi @ggreif 

[...]

PR #540, opened 2020-10-22
Add note about VSCodium update API to "Getting all the Telemetry Out"

tywmick:
Not too long after switching from VS Code, I saw a software update available within VSCodium and downloaded and applied the update without thinking. Later, I was perusing [Getting all the Telemetry Out](https://github.com/VSCodium/vscodium/blob/5ffd1aa4b0f59d51d37be6eeb0d8830bd6c42a40/DOCS.md#getting-all-the-telemetry-out), learned about the `@tag:usesOnlineServices` settings filter, tried using the filter in my own settings, and saw in the description for the `update.mode` setting, "The updates are fetched from a Microsoft online service."

That very much confused me, since I thought it meant that when I had used the internal software update function in VSCodium it had actually fetched those updates from Microsoft instead of from this repository. I remained confused until I happened upon #533, which led me to the [update-api](https://github.com/VSCodium/update-api) repo, which pointed me to #41 and the [Auto Update Support](https://github.com/VSCodium/vscodium/projects/1) project, and I finally understood.

So this PR is a little addition to "Getting all the Telemetry Out" that explains the update API in light of the `update.mode` setting description to hopefully prevent further confusion.

Here's the new text, starting with a new subheading within the "Getting all the Telemetry Out" section:

> ### VSCodium Update API
>
> When searching the `@tag:usesOnlineServices` filter, note that while the "Update: Mode" setting description still says "The updates are fetched from a Microsoft online service", VSCodium's [build script sets](https://github.com/VSCodium/vscodium/blob/master/prepare_vscode.sh#L36) the `updateUrl` field in `product.json` to that of VSCodium's own small [update server](https://github.com/VSCodium/update-api), so enabling that setting won't actually result in any calls to Microsoft servers.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -26,9 +26,11 @@ It is also highly recommended that you review all the settings that "use online 
 
 These can all be disabled.
 
-__Please note that some extensions send telemetry data to Microsoft as well. We have no control over this and can only recommend removing the extension.__
+__Please note that some extensions send telemetry data to Microsoft as well. We have no control over this and can only recommend removing the extension.__ _(For example, the C# extension `ms-vscode.csharp` sends tracking data to Microsoft.)_
 
-_(For example the C# extension `ms-vscode.csharp` sends tracking data to Microsoft.)_
+### VSCodium Update API
+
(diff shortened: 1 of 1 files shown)


stripedpajamas:
awesome, thank you!!

PR #553, opened 2020-10-30
Document more Microsoft online service replacements in "Getting all the Telemetry Out"

tywmick:
Following up on #540, I found another couple of `@tag:usesOnlineServices` settings whose descriptions incorrectly (in VSCodium) state, "The updates are fetched from a Microsoft online service": `extensions.autoCheckUpdates` and `extensions.autoUpdate`. Figured I may as well add a similar note about these settings.

Here's the new version of the section, the name of which I've changed from "VSCodium Update API" to "Replacements for Microsoft Online Services":

> ### Replacements to Microsoft Online Services
>
> When searching the `@tag:usesOnlineServices` filter, note that while the "Update: Mode" setting description still says "The updates are fetched from a Microsoft online service", VSCodium's build script [sets the `updateUrl` field](https://github.com/VSCodium/vscodium/blob/master/prepare_vscode.sh#L36) in `product.json` to that of VSCodium's own small [update server](https://github.com/VSCodium/update-api), so enabling that setting won't actually result in any calls to Microsoft servers.
>
> Likewise, while the descriptions for "Extensions: Auto Check Updates" and "Extensions: Auto Update" include the same phrase, VSCodium [replaces](https://github.com/VSCodium/vscodium/blob/master/prepare_vscode.sh#L42) the Visual Studio Marketplace with Open VSX, so these settings won't call Microsoft, either.

-----

Also, while I was at it, I adjusted the formatting of the `extensionsGallery` JSON code block a little farther down to give it syntax highlighting and align it with the list item it accompanies.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -28,7 +28,7 @@ These can all be disabled.
 
 __Please note that some extensions send telemetry data to Microsoft as well. We have no control over this and can only recommend removing the extension.__ _(For example, the C# extension `ms-vscode.csharp` sends tracking data to Microsoft.)_
 
-### Microsoft online service replacements
+### Replacements to Microsoft Online Services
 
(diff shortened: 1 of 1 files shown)


stripedpajamas:
Thanks for the quality PRs lately @tywmick 

Really appreciate it. This looks dope

tywmick:
@stripedpajamas My pleasure! 😊

PR #565, opened 2020-11-14
update AUR maintainer

binex-dsk:
narcissism at its finest :) 

CRKatri disowned the package for some reason so I'm currently maintaining it.

And seriously, does anyone know why he disowned? I haven't been able to contact him and was wondering if anyone, you know, actually knew.


binex-dsk:
found out why he disowned:

> I disowned cause I no longer use Arch, I just switched to FreeBSD, before that I was using void so I've been a bit slow before but now I don't even have pacman to test

stripedpajamas (approved):

PR #568, opened 2020-11-15
Add more proposed APIs

proletarius101:
*   C++ tools
*   Python
*   Docker
*   Intellicode

Not all functions can be enabled by this subtle modification. For instance, you still can't use pylance.

### DIFF
--- a/prepare_vscode.sh
+++ b/prepare_vscode.sh
@@ -53,7 +53,7 @@ win32AppUserModelId='setpath(["win32AppUserModelId"]; "Microsoft.VSCodium")'
 win32ShellNameShort='setpath(["win32ShellNameShort"]; "VSCodium")'
 win32x64UserAppId='setpath (["win32x64UserAppId"]; "{{2E1F05D1-C245-4562-81EE-28188DB6FD17}")'
 urlProtocol='setpath(["urlProtocol"]; "vscodium")'
-extensionAllowedProposedApi='setpath(["extensionAllowedProposedApi"]; getpath(["extensionAllowedProposedApi"]) + ["ms-vsliveshare.vsliveshare", "ms-vscode-remote.remote-ssh"])'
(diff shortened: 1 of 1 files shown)


stripedpajamas (approved):

Kein:
> For instance, you still can't use pylance.

Any ideas what can be hacked together to enable it?

proletarius101:
> > For instance, you still can't use pylance.
> 
> Any ideas what can be hacked together to enable it?

Sadly not yet.

Kein:
Do you know what exactly the issue that prevents it from working?

proletarius101:
Well I don't think it's easy to reverse engineering or something. But probably it's some code in `dist/extension.bundle.js`. I would recommand https://github.com/microsoft/pyright instead

Kein:
Pylance is Pyright, it is the front version for VSCode.

proletarius101:
How about Python with Microsoft language server + Pyright? Then you'll have IntelliCode. Does that works well? The down side I can see is in Jupyter Pyright will not work.

Anyway, it's regrettable that they choose not to open source pylance and not to integrate Pyright with vscode-python. But unfortunately that's their investment. There is no way to force them to open source. (IntelliCode is no more than LSTM at least for Python... Perhaps you can even just decode their onnx model... That may not be copyrighted.

Kein:
> How about Python with Microsoft language server + Pyright?

Yea, that's what I meant when I said "Pyright", it just their version of mypy, essentially. Uniques of Pylance is that it offer features for python no other project/toolkit offers. Which is a characteristic of evil approach MS took when they """""went open-source"""" - they either outright buy community projects (when they killed Omisharp for sublime) or, like with Pyright/Oylance, instead of adopting and improving existing solutions, they reinvent their own wheel, add unique features only big corp with infinite budget can (by throwing 9k coders at it) and now they can justify the value of VSCode, which in turn, will provide its own value with private data hoarding from users.

It is ingenious, really.

PR #569, opened 2020-11-15
Don't replace product.json in RPM updates

proletarius101:
If product.json is modified locally, say, to change gallery url, in each updates, the product.json is replaced by the vanilla one, which makes some troubles. This patch prevents that issue in RPM builds by marking `noreplace` on it. Not sure whether other packaging formats have similiar feature.

### DIFF
--- a/patches/no-replace-product-json.patch
+++ b/patches/no-replace-product-json.patch
@@ -1,0 +1,8 @@
+--- vscode/resources/linux/rpm/code.spec.template	2020-11-15 15:28:20.179070106 +0800
++++ vscode/resources/linux/rpm/code.spec.template.new	2020-11-15 15:25:39.269000000 +0800
+@@ -69,3 +69,5 @@
+ /usr/share/pixmaps/@@ICON@@.png
+ /usr/share/bash-completion/completions/@@NAME@@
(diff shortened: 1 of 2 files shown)


stripedpajamas (approved):

stripedpajamas:
Cool idea :D thank you for doing this.

GitMensch:
I suggest to revert this change given the new features to both set the commonly requested changes of the marketplace via environment variables or with a user specific product json that only is used in a "diff" approach that are now available because of https://github.com/VSCodium/vscodium/pull/674.
product.json also includes a bunch of other stuff than the marketplace and most of those changes should be taken over after an RPM update.

daiyam:
I agree 100% with @GitMensch

proletarius101:
Yeah, If we don't need to change product.json actively, it becomes useless and causes problems

proletarius101:
Wait... Maybe we also need to change the `proposedApi`?

proletarius101:
As a side note, it seems possible to add a user-wide `product.json`, which solves all issues: https://github.com/VSCodium/vscodium/blob/master/DOCS.md#how-to-use-the-vs-code-marketplace

Issue #584, opened 2020-12-12
There is no macOS release for 1.52.0


hazarek:
**Describe the bug**
There is no macOS release for 1.52.0
**Please confirm that this problem is VSCodium-specific**
- [x] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

**Please confirm that the issue/resolution isn't already documented**
- [x] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**To Reproduce**
1. Go to https://github.com/VSCodium/vscodium/releases/tag/1.52.0
2. No macos release

---
1. brew upgrade --cask vscodium
2. No upgrade
---
1. VScodium menu > Check for Updates
2. "There are currently no updates available."

---
**Desktop (please complete the following information):**
 - OS: macOS 11.0.1
 - Architecture intel
 - Version: VScodium 1.51.1

hazarek:
Also previous release file name is incomplete. OS name is not written.
https://github.com/VSCodium/vscodium/releases/download/1.51.1/VSCodium.1.51.1.dmg

`VSCodium.1.51.1.dmg` -> `VSCodium-macos-1.51.1.dmg`

MPLew-is:
Looks like this is probably due to the macOS build step failing: https://github.com/VSCodium/vscodium/runs/1540496947?check_suite_focus=true

hazarek:
i cant wait this bullshit, i will switch to vscode.

revta:
@MPLew-is That would probably explain why the latest build cant be opened on macOS. I'm on 10.14.6 and mac wont let it be opened, complaining about the signature.... That would probably explain why...

MentalGear:
Seems like macOS build are still not generated.

MentalGear:
Thx, took me a while to find it as I expected a "mac" identifier, similar to the used "win" and "linux" ones.

PR #586, opened 2020-12-14
Include arch in Darwin build task

stripedpajamas:
MS looking to support Mac ARM builds, so the gulp task was renamed.

Ref:
- https://github.com/microsoft/vscode/blob/master/build/azure-pipelines/darwin/product-build-darwin.yml#L141
- https://github.com/microsoft/vscode/blob/master/build/azure-pipelines/product-build.yml#L167


stripedpajamas:
Closes #584

PR #594, opened 2020-12-23
fix check-tags.sh on macos

daiyam:
This PR fixes `check-tags.sh` where the architecture type in the archive name is missing.


stripedpajamas (approved):

stripedpajamas:
Good catch :D Thank you

PR #595, opened 2020-12-23
fix update_version.sh on macos

daiyam:
This PR adds `VSCODE_ARCH` to the archive and to `VERSION_PATH ` but not sure if it's correct...

### DIFF
--- a/update_version.sh
+++ b/update_version.sh
@@ -99,7 +99,7 @@ cd ..
 
 if [[ "$OS_NAME" == "osx" ]]; then
   # zip, sha1, and sha256 files are all at top level dir
-  ASSET_NAME=MrCode-darwin-${VSCODE_ARCH}-${LATEST_MS_TAG}.zip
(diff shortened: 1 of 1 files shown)


update_version.sh
```diff
@@ -99,8 +99,8 @@ cd ..
 
 if [[ "$OS_NAME" == "osx" ]]; then
   # zip, sha1, and sha256 files are all at top level dir
-  ASSET_NAME=VSCodium-darwin-${LATEST_MS_TAG}.zip
-  VERSION_PATH="darwin"
+  ASSET_NAME=MrCode-darwin-${VSCODE_ARCH}-${LATEST_MS_TAG}.zip
```
stripedpajamas, line 102: Should be VSCodium not MrCode

daiyam: Oops... fixed

daiyam:
@stripedpajamas I've updated `update-api` to add a default arch (x64) for macos. So now, this PR is correct.

stripedpajamas (approved):
Another good catch :D thank you

PR #649, opened 2021-02-22
supported platform update, fixes #648

### LINKED ISSUE #648: What is the minimal version of Mac OS X to run VSCodium?
Visual Studio Code lists Mac OS X 10.10+ - is it the same for VSCodium? Did I miss some documentation or is this information missing in the docs currently?

Could VSCodium go back more (possibly at least with a manual build, when using the binary in homebrew needs a newer OS)?

GitMensch:
* adds some minimal versions, along with the reason for it (at least I was tempted to say "we could go further than Visual Studio Code" - but we actually can't because of Electron
* adds arm64 macOS status
* adds Windows arm64 which is actually provided since some time

Note: I highly suggest to change "supported OS" to "supported platforms" as "Linux" is no OS but a kernel and we explicit mention the processors, too (I personally also think GNU/Linux would be better in this context).

If approved I'd also do both these changes and force-push so we have a single commit for this @stripedpajamas your decision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@
   - [Flatpak Option](#flatpak)
 - [Why Does This Exist](#why)
 - [More Info](#more-info)
-- [Supported OS](#supported-os)
+- [Supported Platforms](#supported-platforms)
 
 ## <a id="download-install"></a>Download/Install
 :tada: :tada: [Download latest release here](https://github.com/VSCodium/vscodium/releases) :tada: :tada:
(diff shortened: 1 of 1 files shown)


stripedpajamas:
agreed, let's make it "platforms" and GNU/Linux

stripedpajamas (approved):

GitMensch:
pushed the additional changes, @stripedpajamas have fun inspecting/merging :-)

stripedpajamas (approved):

stripedpajamas:
Thanks @GitMensch

PR #650, opened 2021-02-23
Remove redundant `yarn postinstall` call

Livven:
`yarn postinstall` is automatically run after `yarn install`, so this explicit call is not necessary.

Removing it saves around 10 to 20 seconds in the build job ([Linux](https://github.com/VSCodium/vscodium/runs/1883768900?check_suite_focus=true#step:8:597), [macOS](https://github.com/VSCodium/vscodium/runs/1883686772?check_suite_focus=true#step:6:579), [Windows](https://github.com/VSCodium/vscodium/runs/1883728672?check_suite_focus=true#step:8:589) logs for reference).


stripedpajamas (approved):

stripedpajamas:
Awesome, thanks @Livven

Issue #644, opened 2021-02-17
Users can accidentaly install non-functional Codium due missing dependency information on older enterprise distributions e.g.CentOS7


Sinihopea:
Updating into the latest Codium RPM will leave program in partially non-functional state due changed CXXABI requirements in upstream. These dependencies haven't been properly specified in the RPM .spec file and so doing e.g. "yum upgrade" will suddenly cause Codium to have some major unexpected issues. I just encountered this problem yesterday having updated my workstation.

RHEL7 and CentOS7 do not and will not offer the newer CXXABI newer Codium requires. This stems from upstream change in VSCode itself where they moved to Electron Framework 11 which has a dependency to newer C++ libraries. The lack of these newer libraries leads to for example non-functional terminal within Codium and several extensions breaking.

Their changelog notice about this: https://code.visualstudio.com/updates/v1_53#_electron-11-update

Upstream has had several issues about this which are closed because it's "won't fix" issue. Only thing users can do is to upgrade into newer C++ toolchain or downgrade the editor itself. For example some of these threads:

- https://github.com/microsoft/vscode/issues/115784
- https://github.com/microsoft/vscode/issues/115851
- https://github.com/microsoft/vscode/issues/115871

I downgraded and added lock into Codium and things work now. I fully understand and do not expect Codium to receive any longer updates which will work on CentOS7. However it would be nice if you could add into RPM .spec file that Codium now requires this specific level of CXXABI to fully function.

Like mentioned by this user: https://github.com/microsoft/vscode/issues/115784#issuecomment-774170403

Thank you.

daiyam:
Can you do a PR?

Sinihopea:
Eh? No. I'm an end-user of Codium who stumbled into this issue. I don't know if they've updated the RPM .spec in VSCode upstream. I mostly just wanted to bring this bug to your attention because current Codium "el7" package isn't actually compatible with el7 because of this dependency issue.

paulcarroty:
[The workaround](https://github.com/microsoft/vscode/issues/115784#issuecomment-781031750)

VSCode (and VSCcodium too) now use Electron 11 & Ubuntu 18.04, so old distros will get problems sooner or later. Solution: upgrade to new release or use/create the backports.

Sinihopea:
> [The workaround](https://github.com/microsoft/vscode/issues/115784#issuecomment-781031750)
> 
[...]

PR #660, opened 2021-03-02
remove Centos 7 compatibility

daiyam:
This PR is removing the compatibility for Centos 7 since it isn't supported by Electron 11 (https://code.visualstudio.com/updates/v1_53#_electron-11-update).

### DIFF
--- a/patches/fix-rpm-spec.patch
+++ b/patches/fix-rpm-spec.patch
@@ -1,0 +1,16 @@
+--- vscode/resources/linux/rpm/code.spec.template	2021-03-02 19:26:53.000000000 +0100
++++ vscode/resources/linux/rpm/code.spec.template.new	2021-03-02 19:28:12.000000000 +0100
+@@ -1,6 +1,6 @@
+ Name:     @@NAME@@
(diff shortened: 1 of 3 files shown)


stripedpajamas (approved):

stripedpajamas:
Thanks @daiyam 👍

GitMensch:
> > Then maybe we should make a patch so the line 3 became `Release: @@RELEASE@@.el8`? Will it be ok?

> Yes, thank you. And while you're at it, please add `Requires: libstdc++.so.6(CXXABI_1.3.9)` so that package managers like yum, dnf, zypper, etc. can tell what this package requires to run at minimum as well and resolve the dependencies on their own. And in the case of CentOS7 and EL7 distributions in-general block the install process since EL7 distros do not support that newer ABI. This is why this issue exists.

_Originally posted by @Sinihopea in https://github.com/VSCodium/vscodium/issues/644#issuecomment-790584988_

So what do you think about further adjusting this? Explicit ask for a compatible libstd6++ sounds quite reasonable.

PR #665, opened 2021-03-08
Reduce context for `patches/update-cache-path.patch`

MPLew-is:
Upstream changed the next line, causing build breakage. Reducing the context resolves the issue and should help prevent future issues if/when that function changes again.

Examples of failing builds this should fix (though I have only tested Linux):
- https://github.com/VSCodium/vscodium/actions/runs/630661540
- https://github.com/VSCodium/vscodium/actions/runs/630646856
- https://github.com/VSCodium/vscodium/actions/runs/630631432

This should unblock releasing 1.54.0 and 1.54.1, which were released upstream late last week.

PR #666, opened 2021-03-10
fixing build

daiyam:
This PR is fixing the build.

Now, every PR are fully building VSCodium and aren't trying the publish it when done. 

### fixing build for 1.54

It's using the PR #665 by @MPLew-is.
And it's fixing a regression in `fix-rpm-spec.patch` by me (#660).

### adding the variable `SHOULD_DEPLOY`.

When the branch is a PR, `SHOULD_BUILD` is set to `yes` and `SHOULD_DEPLOY` to `no`
It isn't a PR, `SHOULD_DEPLOY` is set `yes` and `SHOULD_BUILD` is determined in `check_tags.sh`

### adding cache

This PR is adding the cache for yarn cache. It's avoid network errors which can occurs from time to time.
It's only save from 1 to 2 minutes.

### add scripts to build VSCodium by hand

I've added the scripts `build/build_windows.ps1`, `build/build_macos.sh` and `build/build_linux.sh` so we can easily build VSCodium by hand.


stripedpajamas:
Thanks so much @daiyam, looks good

Also thanks @MPLew-is for the initial work.

stripedpajamas (approved):

MPLew-is:
Looks like this broke the release logic, and new builds aren't generating release assets or tags: https://github.com/VSCodium/vscodium/actions/runs/644566547

daiyam:
:scream::sob::sob::sob:

daiyam:
@MPLew-is Thx, the PR #668 should fix my mistake.

PR #668, opened 2021-03-12
fix schedule runs

daiyam:
This PR is fixing a regression introduced in #666 which blocks the creation of any new release...
It's correctly determining when the run is a PR or not.

### DIFF
--- a/check_cron_or_pr.sh
+++ b/check_cron_or_pr.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-if [[ $github.event.number ]]; then
+if [[ $github.event_name == "pull_request" ]]; then
 	echo "It's a PR"
 	
 	export SHOULD_BUILD="yes"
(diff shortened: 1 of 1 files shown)


check_cron_or_pr.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-if [[ $github.event.number ]]; then
+if [[ $github.event_name == "pull_request" ]]; then
```
MPLew-is, line 5: I think this should be [`GITHUB_EVENT_NAME`](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables) instead, since we're in the shell here.

MPLew-is (changes requested):

check_cron_or_pr.sh
daiyam: Ya, I'm still learning... Thx

MPLew-is (approved):

check_cron_or_pr.sh
daiyam: Thx It's working.

stripedpajamas (approved):

Issue #401, opened 2020-05-14
VSCodium 1.45.0 cannot authenticate with Github


penyuan:
**Describe the bug**
VSCodium 1.45.0 cannot authenticate with Github.

**Please confirm that this problem is VSCodium-specific**
- [x] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

**Please confirm that the issue/resolution isn't already documented**
- [x] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**To Reproduce**
1. Clone a Github repository and work on it, committing changes, work on branches locally, etc.
2. Push to remote or publish a branch, etc.
3. Click "Allow" on the dialog that says "The extension 'Git' wants to sign in using Github."
4. Click "Open" on the dialog that says "Do you want VSCodium to open the external website?"
5. A page launches in my web browser (Firefox 76.0.1) with the header "Authorize Visual Studio Code to access GitHub", click "Continue." (**I wonder if this shows the problem because VSCodium is identifying itself to Github as Visual Studio Code???**)
6. It fails with a page that says "Oh no! An error occurred! Please restart the sign in process from the editor. Forbidden".

**Expected behavior**
Github web authentication to succeed without error.

**Screenshots**
Here is the Github error page:
![](https://framapic.org/UIpNRvRQy5ZI/wwJZFIkjlSuJ)

**Desktop (please complete the following information):**
 - OS: RHEL 8.2 (Gnome 3.32.2 in Xorg mode with Linux kernel 4.18.0)
 - Architecture: x86_64
 - Version: 1.45.0

**Additional context**
This does **not** happen with Microsoft's build of Visual Studio Code.

marcorichetta:
Related => https://github.com/microsoft/vscode-pull-request-github/issues/1778#issue-614883422

Also from the [wiki of the Github PR Extension](https://github.com/microsoft/vscode-pull-request-github/wiki#3-why-isnt-the-extension-starting), although this issue is related with Github Auth.

penyuan:
Thank you @marcorichetta! I've subscribed to the issue #1778 you linked to, looks like it is an ongoing issue.

How can we help fix it?

marcorichetta:
Hi @penyuan.

I tried the solutions on the linked issue but none of them worked for me. Maybe it's related to Github's backend but I don't know really.

### VSCodium
Version: 1.45.0
[...]

PR #672, opened 2021-03-19
use github personal access token

daiyam:
This PR is replacing the cumbersome auth relay server (#401) with github personal access token.

How to create a PAT: https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token

### DIFF
--- a/prepare_vscode.sh
+++ b/prepare_vscode.sh
@@ -10,6 +10,7 @@ cd vscode || exit
 # apply patches
 patch -u src/vs/platform/update/electron-main/updateService.win32.ts -i ../patches/update-cache-path.patch
 patch -u resources/linux/rpm/code.spec.template -i ../patches/fix-rpm-spec.patch
+patch -u extensions/github-authentication/src/githubServer.ts -i ../patches/use-github-pat.patch
 
 if [[ "$OS_NAME" == "osx" ]]; then
   CHILD_CONCURRENCY=1 yarn --frozen-lockfile --ignore-optional
--- a/patches/use-github-pat.patch
+++ b/patches/use-github-pat.patch
@@ -1,0 +1,152 @@
+--- a/extensions/github-authentication/src/githubServer.ts
++++ b/extensions/github-authentication/src/githubServer.ts
+@@ -6,15 +6,12 @@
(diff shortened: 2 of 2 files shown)


pmahend1:
I was looking at the issues for this and came across this. Is it possible to have app level Auth provider without PAT, may be that will help other extensions use the same provider seemlessly without users having to manually enter PAT?

daiyam:
@GitMensch Thx

zyansheep:
What is the holdup on this getting merged?

daiyam:
@stripedpajamas What do you think about this PR?

stripedpajamas:
It feels to me like it will be a pain to maintain this patch as upstream continues development. It's fine to merge it in, just foreseeing it drift like the other patches and this one is more complex imo.

daiyam:
@stripedpajamas This patch isn't so complex. I just remove some parts of the code, nothing more. So it's easy.
I will add a section in the doc about that change, then I will merge it.

prepare_vscode.sh
```diff
@@ -19,7 +19,6 @@ for file in ../patches/*.patch; do
     fi
   fi
 done
-
```
rugk, line 22: Huh? And unintended change I guess?

rugk (changes requested):

prepare_vscode.sh
daiyam: Good catch! Thx

GitMensch:
Looks good to me, I'd also merge after CI is finished. It should definitely be integrated into tonight's 1.60!

daiyam:
@GitMensch Yep. That's why I've made it.

daiyam:
@GitMensch Can you check the doc if it's good enough? Thx

GitMensch:
I already did,and wanted to explicit write "thanks for the doc update", somehow that got lost from my typing...
Yes, docs are fine. Thanks again.

rugk:
FYI, did I mention that GitHub doing diff's on .diff files is absolutely crazy?

Just when you need something to laugh look un _unified_ view at one or few xcommits in this PR: :sweat_smile: 

image

PR #695, opened 2021-04-06
Fix Jupyter extension crash

jingyu9575:
The latest Jupyter extension crashes during start on my system, with the following error messages: 

> Extension activation failed, run the 'Developer: Toggle Developer Tools' command for more information.
> 
> extension activation failed Error: [ms-toolsai.jupyter]: Proposed API is only available when running out of dev or with the following command line switch: --enable-proposed-api ms-toolsai.jupyter

I found that adding `"ms-toolsai.jupyter"` to `extensionAllowedProposedApi` can fix this problem. It is listed in *product.json* of Microsoft's VSCode. BTW there are many more extension IDs in Micorsoft's *product.json* but I'm not sure if copying this field is a good idea.


proletarius101:
> BTW there are many more extension IDs in Micorsoft's product.json but I'm not sure if copying this field is a good idea.

It doesn't hurt and has no legal implication. Besides, since we want to revert https://github.com/VSCodium/vscodium/pull/569, it's a good idea to put all those API IDs into ours.

jingyu9575:
@proletarius101 I've added all the IDs from VSCode 1.55.1 in the commit.

These IDs: `"ms-vscode.cpptools", "ms-azuretools.vscode-docker", "visualstudioexptteam.vscodeintellicode"` were in VSCodium but not in VSCode. I have kept them at the beginning.

IceWreck:
I added it to my product.json as well but its not fixing the crash.

prassee:
Same here I've moved to VSCode just to get rid of this error. Pyright also impacted due to this error.

PR #730, opened 2021-06-04
Vscodium officially available in Gentoo repos

papjul:
No longer requires to install an overlay.

See https://packages.gentoo.org/packages/app-editors/vscodium


stripedpajamas:
I don't know much about Gentoo -- should the section be replaced with instructions on installing it from the official repo?

papjul:
Installing a package through the package manager is really a basic ;)

If you want to add it, it’s `emerge vscodium` but it would be more relevant in a section with instructions for other Linux distributions or OS in that case.

stripedpajamas (approved):

Issue #330, opened 2020-02-14
AppImage Crashes on open dialog


rpangrazio:
When using the AppImage, VSCromium opens fine, but then once i go to open a folder it crashes. Here is the console output

libGL error: failed to open drm device: Permission denied
libGL error: failed to load driver: vmwgfx
Warning: 'sandbox' is not in the list of known options, but still passed to Elec
tron/Chromium.
[main 2020-02-14T13:04:52.617Z] update#setState idle
bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
Error creating proxy: Unknown or unsupported transport “disabled” for address “d
isabled:” (g-io-error-quark, 13)

(codium:106554): Gtk-WARNING **: 07:04:57.477: Could not find the icon 'user-hom
e-symbolic-ltr'. The 'hicolor' theme
was not found either, perhaps you need to install it.
You can get a copy from:
        http://icon-theme.freedesktop.org/releases

(codium:106554): Gtk-WARNING **: 07:04:57.481: Could not load a pixbuf from /org
/gtk/libgtk/icons/16x16/status/image-missing.png.
This may indicate that pixbuf loaders or the mime database could not be found.
**
Gtk:ERROR:../../../../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertio
n failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/imag
e-missing.png: Unrecognized image file format (gdk-pixbuf-error-quark, 3)
Aborted (core dumped)

OS Information:
Ubuntu 18.04 server inside VMWare VM
uname -a:
Linux bionic 4.15.0-76-generic #86-Ubuntu SMP Fri Jan 17 17:24:28 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

Other things to note:
I run CDE as a desktop UI.

I have tried both 1.42.0 and 1.42.1

srevinsaju:
Reproducible on Arch Linux

lvarin:
Similar, if not the same:

```
[main 2020-10-07T06:15:32.521Z] update#checkForUpdates, state =  idle
[main 2020-10-07T06:15:32.521Z] update#setState checking for updates
[main 2020-10-07T06:15:32.523Z] update#setState idle
[main 2020-10-07T06:15:53.866Z] menubarService#updateMenubar 1
[main 2020-10-07T06:15:54.818Z] menubarService#updateMenubar 1
[...]

Issue #532, opened 2020-10-09
AppImage fails to create in Docker environment


stripedpajamas:
Recently switched to building in containers, the same as Microsoft is doing. Now the AppImage [fails to build](https://github.com/stripedpajamas/vscodium/runs/1229304191?check_suite_focus=true) with this message:

```
fuse: device not found, try 'modprobe fuse' first

Cannot mount AppImage, please check your FUSE setup.
You might still be able to extract the contents of this AppImage 
if you run it with the --appimage-extract option. 
See https://github.com/AppImage/AppImageKit/wiki/FUSE 
for more information
open dir error: No such file or directory
Error: Process completed with exit code 127.
```

This seems to be a known issue (https://github.com/AppImage/AppImageKit/issues/405, https://github.com/AppImage/AppImageKit/wiki/FUSE#docker), but I'm not sure exactly how to fix it. Looking for help from perhaps @linsui who got the AppImage scripts in our build.

linsui:
I'd like to help but I don't have enough skill... Forward ping to @tyu1996 :joy:

stripedpajamas:
Should be fixed in next release, thanks @ferion11

PR #731, opened 2021-06-05
linux: fix appimage

daiyam:
This PR is replacing the old `pkg2appimage` binary with its latest version.
It's also generating the `zsync` file so that the appimage becomes updateable.

It's fixing the issue #330.


create_appimage.sh
```diff
@@ -1,9 +1,18 @@
 #!/bin/bash
+
+cd ..
+
 if [[ "$VSCODE_ARCH" == "x64" ]]; then
   # install a dep needed for this process
   sudo apt-get install desktop-file-utils
 
-  cd ..
-  export DOCKER_BUILD=1
```
stripedpajamas, line 7: Does the latest version of the appimage utility auto detect it's running in a docker container now? This was required before.

daiyam: It seems so with [`grep docker /proc/1/cgroup >/dev/null && export DOCKER_BUILD=1 || true`](https://github.com/AppImage/pkg2appimage/blob/23e510b394fc28d24386e2eba076132a646f456c/functions.sh#L33).

But I'm having an issue the imagemagick's `convert` which are failing the packaging. And the error `Cannot mount AppImage, please check your FUSE setup.` is weird...

stripedpajamas:
See https://github.com/VSCodium/vscodium/issues/532 for what the issue was last time

daiyam:
OK, an AppImage is launched to make the AppImage. So I understand the error and know what to test. Thx

daiyam:
The issue was that `convert` is also provided by `pkg2appimage` but as an AppImage which fail when called in Docker...
So I've replaced the path to use the system's `convert` which is a regular binary file.

stripedpajamas (approved):

PR #765, opened 2021-07-16
Use `applicationName` for update cache path

Livven:
Seems more appropriate, since `nameShort` is intended to be user-facing and could e.g. contain spaces.

### DIFF
--- a/patches/update-cache-path.patch
+++ b/patches/update-cache-path.patch
@@ -7,7 +7,7 @@ index 93424ca..133657e 100644
  	@memoize
  	get cachePath(): Promise<string> {
 -		const result = path.join(tmpdir(), `vscode-update-${this.productService.target}-${process.arch}`);
-+		const result = path.join(tmpdir(), `${this.productService.nameShort.toLowerCase()}-update-${this.productService.target}-${process.arch}`);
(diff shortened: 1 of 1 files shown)


stripedpajamas:
This makes sense to me. Will want @daiyam to chime in since I think he authored the patch originally.

stripedpajamas (approved):

patches/update-cache-path.patch
```diff
@@ -7,7 +7,7 @@ index 93424ca..133657e 100644
  	@memoize
  	get cachePath(): Promise<string> {
 -		const result = path.join(tmpdir(), `vscode-update-${this.productService.target}-${process.arch}`);
-+		const result = path.join(tmpdir(), `${this.productService.nameShort.toLowerCase()}-update-${this.productService.target}-${process.arch}`);
++		const result = path.join(tmpdir(), `${this.productService.applicationName}-update-${this.productService.target}-${process.arch}`);
```
stripedpajamas, line 10: I think we'd still want to keep the `toLowerCase` though, for consistency sake

daiyam: I think `this.productService.applicationName` is fine as it is since it's also the name of the executable (without any lower casing).
But yeah, the code is better than the one I did...

daiyam (approved):

Issue #763, opened 2021-07-13
Different icon for different platform


pmahend1:
**Describe the bug**

- Windows:  *blue tree with circular white background*
- Mac :   *blue tree with white background*
- Ubuntu: *has old green icon*

**Please confirm that this problem is VSCodium-specific**
- [x] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

**Please confirm that the issue/resolution isn't already documented**
- [x] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**To Reproduce**
- Install VSCodium on Windows
- Install VSCodium on Mac
- Install VSCodium on Ubuntu
- 
**Expected behavior**

Same icon should be used on all platforms.

**Screenshots**

**Desktop (please complete the following information):**
 - OS:Windows 10, Mac BigSur , Ubuntu 20
 - Architecture : x64
 - Version : 1.58.0

**Additional context**
NA

hazarek:
i don't want to use white background on mac.

pmahend1:
@hazarek 

> 
> 
> i don't want to use white background on mac.

Icon should be consistent , can you submit a PR to replace Windows and Linux icons with the non-white background one?

hazarek:
> @hazarek
> 
> > i don't want to use white background on mac.
> 
> Icon should be consistent , can you submit a PR to replace Windows and Linux icons with the non-white background one?

I can only send as png or icns. I don't have technical knowledge about linux and windows icons.

daiyam:
I agree that the icons need to be the same. I will make the necessary changes.

daiyam:
Let's keep it active.

aflip:
On pop OS, the codium icon on the taskbar is blank. Is this due to the same issue? and no icon is found in the applications menu.

explode:
The logo on popup dialog boxes looks weird - like an SVG path not rendering properly.  Main app logo also shown (for side-by-side comparison).  Kubuntu 22.04, Linux kernel 5.15.0-72, Qt version: 5.15.3, Graphics platform: X11
![badly rendered logo](https://github.com/VSCodium/vscodium/assets/47583746/358826c3-8012-464c-b3f1-bb534473098d)

andre-koch:
I can't figure it out either. on Kali it shows the correct icon but in Ubuntu its incorrect!

Issue #591, opened 2020-12-20
[VOTE] macOS BigSur app icon update


hazarek:
Will need a vote for big sur icons; will merge this as is for now.
_Originally posted by @stripedpajamas in https://github.com/VSCodium/vscodium/issues/582#issuecomment-748263074_

## Before/After
<img width="187" alt="codium_vote" src="https://user-images.githubusercontent.com/36481442/102703049-14a99200-427b-11eb-937c-416055f4b746.png">

I recommend the optimized version, there will be no visual difference.

ICNS file (629KB  1024x1024 144dpi Multiple size);
https://yadi.sk/d/mIlIdbAL2_gqpg

ICNS file (126KB 512x512 72dpi Optimized version);
https://yadi.sk/d/POI_637hLAgF0g

Yes 👍🏼 
No 👎🏼

hazarek:
Also document icons need to be updated for Big Sur. if you want i can redesign and update them all.
https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/document-icons/

hazarek:
@stripedpajamas The result of the voting is clear. It should be included in the next VScodium build so as not to wait long. If you are not sure about the result, you can pin it on the "Pinned issues"

R0LA1mRifcF8yAkk:
Sorry, how do you mean Yes or No? In reference to what? Yes to what? No to what?

stripedpajamas:
@R0LA1mRifcF8yAkk Yes (:+1:) is a vote to change the VSCodium logo (for Mac users) to the blue square icon in the picture. No (:-1:) is a vote to leave the logo as is (white circle). Note that the latest release still has a green logo but that will change upon next release to logo at the top of the README.

hazarek:
> Sorry, how do you mean Yes or No? In reference to what? Yes to what? No to what?

👎🏼 keep the current icon for macOS Big Sur.
👍🏼 I want the new icon at the bottom for macOS Big Sur.

R0LA1mRifcF8yAkk:
Right. Could there be more than just 1 alternative? How about one with white background but instead of the circle it'd have a square?

The blue background is an almost entirely different visual. Might be difficult to identify or get used to when OS-hopping.

hazarek:
white is too bright for dark mode. but blue is not too dark for light mode. are you designer?

R0LA1mRifcF8yAkk:
[...]

PR #768, opened 2021-07-20
New light-colored macOS icon, consistent with logo on other platforms

morellam-dev:
Related to #763. I think the current macOS icon is an interesting design, but I don't like how different it is from the icons on other platforms, and I personally prefer the lighter colors from the Linux/Windows icon. With that in mind, I created this alternative which is identical, except the background is a rounded rectangle.

Current icon:
Current VSCodium icon

Alternative icon:
Redesigned VSCodium icon

This PR simply replaces src/resources/darwin/code.icns with this new icon, scaled down to 512px and converted using libicns.


pmahend1 (changes requested):
I think you missed few places , for Linux .
I really dont know which ones are used for Linux, except  `letterpress*` are not obviously used. On Ubuntu it shows as green icon on its app store .

```
vscodium\icons\codium_only.svg
vscodium\src\src\vs\workbench\browser\media\code-icon.svg
vscodium\src\src\vs\workbench\browser\parts\editor\media\letterpress-dark.svg
vscodium\src\src\vs\workbench\browser\parts\editor\media\letterpress-hc.svg
vscodium\src\src\vs\workbench\browser\parts\editor\media\letterpress.svg
vscodium\src\resources\linux\code.png
vscodium\src\resources\linux\rpm\code.xpm
```

daiyam:
The following files must not be updated.
```
vscodium\icons\codium_only.svg
vscodium\src\src\vs\workbench\browser\parts\editor\media\letterpress-dark.svg
vscodium\src\src\vs\workbench\browser\parts\editor\media\letterpress-hc.svg
vscodium\src\src\vs\workbench\browser\parts\editor\media\letterpress.svg
```

hazarek:
Voting was held and the blue one was chosen. You cannot change it without voting again.

pmahend1:
@hazarek Could you please link the voting results link here?

morellam-dev:
@pmahend1 Presumably referring to the vote here: #591. I’m only a user of this project, so I wasn’t aware of that discussion till now.

As @h0adp0re said then, the only choices presented were 👍 (use the dark icon, which follows the Big Sur “rounded square” guidelines), or 👎(keep the circular light icon, which doesn’t). No option was presented to simply expand the light icon into a rounded square, so that’s what I’ve done here.

With all due respect @hazarek, I understand that you designed this icon, and are quite attached to keeping it. But as a user on several platforms, I would much rather if macOS used the “official” Codium icon, for consistency’s sake. You’re free to use whatever design you want on your own system, but it feels natural to use what is inarguably the “default” design by default.

I’m interested in what @stripedpajamas thinks of this. Is this something that should be put to another popular vote?

pmahend1:
Thank @morellam-dev 

I personally prefer the blue back-grounded icon . But providing just 👍🏻 and 👎🏻  to vote without much explanation seems like incorrect way to poll. People tend to use 👍🏻 since they are more familiar with that , people dont usually use 👎🏻 unless they really want to criticize something on online. Could have used a third party tool like google forms or something.

If blue backgrounded is preferred then Windows and Linux icons need to be updated.

hazarek:
@morellam-dev 
This is an issue about macOS dark and light mode. Icon with white background causes eye strain on macOS dark mode. Also the contrast is very low. Also there is no official logo. This is an open source project. I designed it to match both Dark and Light mode. You should read the details in the related topic.

The two can change dynamically when macOS supports adaptive icons. I don't know if this is possible right now;
https://stackoverflow.com/questions/55529976/macos-mojave-change-app-icon-to-match-dark-light-mode

daiyam:
@morellam-dev Are you using a dark theme/mode? Is this why you would like a lighter icon?

morellam-dev:
@daiyam I sometimes use macOS in dark mode, but it's not really the lightness or darkness that bothers me. It's just that the aesthetics (colors, drop shadows, gradient direction etc.) in the macOS icon are so distinct from the icon in the Windows/Linux builds that it feels disharmonious. I use all three builds pretty frequently, and I just happen to prefer the "white and cyan" version.

From https://github.com/VSCodium/vscodium/issues/4#issuecomment-427070331, it looks like jaredreich actually created a version of the original blue algae icon which has the colors inverted. I mocked up some Big Sur-style derivatives of those (second row), and also a version with the colors darkened to improve contrast, inspired by Hazarek's design (third row). I actually quite like these designs, so I wonder if it's the drop shadow under the emblem that's not appealing to me?

|           | Lite                                                                                                                          | Dark                                                                                                                         |
|-----------|-------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
| Icons by jaredreich  https://github.com/VSCodium/vscodium/issues/4#issuecomment-427070331 | rounded light codium icon  | rounded dark codium icon |
| Converted to rounded rectangle  | square light cyan codium icon  | square dark cyan codium icon |
| Darker blue version |  |  |

Files here: [codium-sur-icons.zip](https://github.com/VSCodium/vscodium/files/7274918/codium-sur-icons.zip). At the very least, these might be useful for people to use as alternative icons on Big Sur.

pmahend1:
If white icon causes strain and not conform to accessibility thresholds then there is no need to keep it on Linux and Windows.
@hazarek Right? 
I personally use VSCodium on both Windows and Mac everyday, both are set on Dark mode. So having two different icons is not consistent. I believe icon should be theme agnostic. Personally using white back-grounded icon is flashy on dark mode for me as well. 
@morellam-dev  I also suggest adding border of 1-2px thickness of different color than the background, to help differentiate if the OS toolbar is pure white or pure blue.

@daiyam IMO its good to have another poll to accommodate opinions

hazarek:
> If white icon causes strain and not conform to accessibility thresholds then there is no need to keep it on Linux and Windows. @hazarek Right? I personally use VSCodium on both Windows and Mac everyday, both are set on Dark mode. So having two different icons is not consistent. I believe icon should be theme agnostic. Personally using white back-grounded icon is flashy on dark mode for me as well. @morellam-dev I also suggest adding border of 1-2px thickness of different color than the background, to help differentiate if the OS toolbar is pure white or pure blue.
> 
> @daiyam IMO its good to have another poll to accommodate opinions

I think the icons on the right are suitable for windows, linux and mac. I think the official icon should be changed to white on blue. I have no problem with the blue ones.

aphive:
I am just now learning of this project, I like the light color background icon but I think that the design itself is way too busy with all the little bulby ends.

stech11845:
> I think you missed few places , for Linux . I really dont know which ones are used for Linux, except `letterpress*` are not obviously used. On Ubuntu it shows as green icon on its app store .
> 
> ```
> vscodium\icons\codium_only.svg
> vscodium\src\src\vs\workbench\browser\media\code-icon.svg
> vscodium\src\src\vs\workbench\browser\parts\editor\media\letterpress-dark.svg
> vscodium\src\src\vs\workbench\browser\parts\editor\media\letterpress-hc.svg
> vscodium\src\src\vs\workbench\browser\parts\editor\media\letterpress.svg
> vscodium\src\resources\linux\code.png
> vscodium\src\resources\linux\rpm\code.xpm
> ```

We are not talking about Linux here, we're talking about MacOS.

GitMensch:
That should be solved together with https://github.com/VSCodium/vscodium/discussions/1218, which also includes the next, more fine grained voting (I _think_ we're still in the discussion phase), no?
Would there be a reason to have a separate variant for MacOS and/or should there be a light and dark mode version in general (possibly best to be discussed in the related issue)?

daiyam:
@GitMensch They should be same on all platform.

GitMensch:
With or without a secondary set for dark mode?

daiyam:
> With or without a secondary set for dark mode?

A secondary set if it can be done.

dionsaur84:
-1

i prefer the current icon

tired of all these white background icons for everything

daiyam:
Closing due to:
- https://github.com/VSCodium/vscodium/discussions/1429
- https://github.com/VSCodium/vscodium/discussions/1430

PR #770, opened 2021-07-21
Add support to use GNU sed installed in mac osx

swinder0161:


swinder0161:
@daiyam Please help to review, this is required is sed is installed from macports or homebrew in mac osx

daiyam:
Can you use the function `is_gnu_sed`  from `https://github.com/VSCodium/vscodium/blob/master/update_settings.sh`? Thx

swinder0161:
@daiyam Wont the function be only visible to the sh file in which it is defined unless export -f is done?

swinder0161:
to get functions at global scope the sh need to be used with source <path to script.sh> but i see the update_settings.sh is called as a sub process
https://github.com/VSCodium/vscodium/blob/master/prepare_vscode.sh#L8

daiyam:
You could move the function in a shared file (like `util.sh`) and import the file with `source "./util.sh"`.

swinder0161:
Ok. All try it tomorrow.

daiyam:
Thx

swinder0161:
Ok, its done now. @daiyam

Issue #746, opened 2021-06-20
[1.58] Extension marketplace not working


swinder0161:
**Describe the bug**
I built my own vscode for macos using main branch of vscode and extension store is not loading
![image](https://user-images.githubusercontent.com/397543/122681047-147d5580-d210-11eb-9735-37e756740402.png)

**Please confirm that this problem is VSCodium-specific**
- [X] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

**Please confirm that the issue/resolution isn't already documented**
- [X] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**To Reproduce**
1. Open VSCodium
2. Click on Extensions button from left side bar
3. See error

**Expected behavior**
Extension store should load

**Screenshots**
![image](https://user-images.githubusercontent.com/397543/122681074-3ecf1300-d210-11eb-9446-2c1ee88debd6.png)

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Architecture x64
 - Version [1.58.0]

**Additional context**
NA

daiyam:
`1.58`??? How have you build the app?

swinder0161:
hi @daiyam 

I apply patch below and run ./build/build_macos.sh

diff --git a/get_repo.sh b/get_repo.sh
index 301efdf..9800f11 100755
--- a/get_repo.sh
+++ b/get_repo.sh
@@ -6,7 +6,7 @@ export LATEST_MS_COMMIT=$(echo $UPDATE_INFO | jq -r '.version')
 export LATEST_MS_TAG=$(echo $UPDATE_INFO | jq -r '.name')
 echo "Got the latest MS tag: ${LATEST_MS_TAG}"

-git clone https://github.com/Microsoft/vscode.git --branch $LATEST_MS_TAG --depth 1
+git clone https://github.com/Microsoft/vscode.git --branch main --depth 1

 # for GH actions
 if [[ $GITHUB_ENV ]]; then

daiyam:
Do you have an error in the Developer Tools (`Help` / `Toggle Developer Tools`)?

swinder0161:
Not an expert at it.
please check screenshot and let me know if I need to provide more details

![image](https://user-images.githubusercontent.com/397543/122728688-0245fa80-d296-11eb-85ac-c879590ed8ff.png)

swinder0161:
[...]

PR #774, opened 2021-07-23
Fix extension marketplace

daiyam:
This PR is fixing #746 by reverting the CORS checks (added in `v1.58`). 

Additionally, it's adding a loop to apply patches.

@stripedpajamas If by the `v1.59.0`, the CORS haven't being added to openvsx, we should apply this PR.

### DIFF
--- a/prepare_vscode.sh
+++ b/prepare_vscode.sh
@@ -8,10 +8,15 @@ cd vscode || exit
 ../update_settings.sh
 
 # apply patches
-patch -u src/vs/platform/update/electron-main/updateService.win32.ts -i ../patches/update-cache-path.patch
-patch -u resources/linux/rpm/code.spec.template -i ../patches/fix-rpm-spec.patch
-git apply --ignore-whitespace ../patches/binary-name.patch
-git apply --ignore-whitespace ../patches/custom-gallery.patch
+for file in ../patches/*.patch; do
+  if [ -f "$file" ]; then
+    echo applying patch: $file;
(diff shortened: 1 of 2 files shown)


prepare_vscode.sh
```diff
@@ -8,10 +8,15 @@ cd vscode || exit
...
-git apply --ignore-whitespace ../patches/custom-gallery.patch
+for file in ../patches/*.patch; do
+  if [ -f "$file" ]; then
+    echo applying patch: $file;
+    if ! git apply --ignore-whitespace $file; then
+      echo failed to apply patch $file 1>&2
+    fi
+  fi
+done
+
```
GitMensch, line 19: That looks like a reasonable thing to include in any case - if you provide that as separate PR and have tested it I can merge that (if you can't do that yet).

daiyam: Yep. I've tested it.

daiyam:
`1.59` should be out by next Wednesday or Thursday.
I don't have any news from OpenVSX team.

GitMensch:
So, shouldn't we close this and have microsoft/vscode#129360 included as patch instead?

daiyam:
@GitMensch I think you have commented on the wrong issue... no?

GitMensch:
Totally.

daiyam:
The `v1.59.0` was tagged Sunday. So the release should in few days.

I will revert the fix when CORS are correctly available on open-vsx.org.

PR #779, opened 2021-07-29
add a loop to apply patches

daiyam:
This PR is adding a loop to apply patches.


daiyam:
@GitMensch here you are :wink:

PR #787, opened 2021-08-05
do not start crashReporter when appCenter isn't configured

daiyam:
The PR is fixing the change between [Electron 13](https://github.com/electron/electron/blob/v13.1.7/docs/api/crash-reporter.md) and [Electron 12](https://github.com/electron/electron/blob/v12.0.13/docs/api/crash-reporter.md) where `submitURL` becomes a requirement.
This is a last minute change found on the `1.59`, I'm pushing the merge before the build process.


daiyam:
That bug could have been a disaster... luckily I've found it. gn

PR #824, opened 2021-09-06
fix: update license menu

daiyam:
This PR is displaying the VSCodium's license from the Help menu.

@stripedpajamas I've updated the license file to include the contributors. I've also added Microsoft because from the Help menu, I think they should be included.
Maybe we should have 2 licence files? 1 for the source code of this repository and 1 for the binary of VSCodium.


stripedpajamas:
This looks fine to me

stripedpajamas (approved):

PR #827, opened 2021-09-08
build: add MSI

daiyam:
This PR is building:
- VSCodium-[Arch]-[Version].msi
- VSCodium-[Arch]-updates-disabled-[Version].msi


icons/build_icons.sh
```diff
@@ -134,6 +134,14 @@ build_win32() {
     convert -size 138x140 xc:white "${SRC_PREFIX}src/resources/win32/inno-small-250.bmp"
     composite -size 116x -gravity center icons/codium_only.svg "${SRC_PREFIX}src/resources/win32/inno-small-250.bmp" "${SRC_PREFIX}src/resources/win32/inno-small-250.bmp"
   fi
+  if [ ! -f "${SRC_PREFIX}build/windows/msi/resources/wix-banner.bmp" ]; then
```
GitMensch, line 137: Both convert and wix (other than Inno) can use PNGs, can you check for an adjustment here to favor PNG over bmp?

GitMensch:
Thanks again for the cleaning and integration of the original PR!

GitMensch:
@daiyam We do have a shell available in the build environment don't we?

I've created a minimal script to convert the license on the fly (works as no RTF special characters are included, if they may be used later `\{}` we need `sed`, if non-ascii 7-bit is used we additionally need either `iconv` and `hexdump` (or add explicit translations via sed for the characters out of that range):

```sh
#!/bin/bash
if test ! -f "$1"; then echo "First argument must specify a file">&2 && exit 1; fi

input=$1
if test "$2" != ""; then target=$2; else target=$1.rtf; fi

cat - >$target <<_EOF
{\rtf1\ansi\deff0{\fonttbl{\f0\fnil\fcharset0 Consolas;}}
\viewkind4\uc1\pard\lang1033\f0\fs22

_EOF

while IFS= read -r line; do
  printf '%s\r\n' "$line"
done  <$input>>$target

printf '}' >>$target
```

That looks a bit easier than the LFS checkout and executable call :-)
What do you think?

daiyam:
Yes, the script is at `build/windows/rtf/make.sh`

GitMensch:
Yes, I remember now. I'll adjust that and remove the executable, then PR.

daiyam:
@GitMensch no PR? :wink:

PR #871, opened 2021-10-12
license conversion txt2rtf

GitMensch:
directly via shell as neither special escaping nor conversion are necessary, removing binaries and LFS from repo and simplify the script


daiyam:
I've never released that rtf was a tex-like... but, also, I've never took a look at them...

PR #900, opened 2021-11-06
retry if file upload fails

MedzikUser:


daiyam:
@MedzikUser Thx! Can you add the check directly into the script ?
Can you also add a delay and the option `--clobber` to `gh release upload` (because it is still creating the files in the release and I need to delete them before re-running the task).

daiyam:
@MedzikUser Sorry, I've made the changes directly. I had enough of the network errors. Thanks again.

PR #960, opened 2021-12-18
feat: build packages only when needed

daiyam:
This PR is avoiding to build packages (like MSI) for every commit.


GitMensch:
That drops the automated builds because the msi folder isn't getting any diffs, no?

PR #970, opened 2022-01-03
remove MacOS depencencies part in build guide

mmvanheusden:
for some reason macos is included in the build guide but with no dependencies? so i removed it


daiyam:
It's there because there are dependencies but it need to be filed in...

PR #985, opened 2022-02-02
Fix invalid `ProductVersion` in insider build

eternalphane:
When building with `INSIDER=1` on windows, WIX will complain with `error LGHT0204: ICE24: ProductVersion: 1.64.0-insider is an invalid version string`. This pr removes the `-insider` suffix from `ProductVersion`.

### DIFF
--- a/build/windows/msi/build.sh
+++ b/build/windows/msi/build.sh
@@ -60,7 +60,7 @@ BuildSetupTranslationTransform() {
 }
 
 "${WIX}bin\\heat.exe" dir "${BINARY_DIR}" -out "Files-${OUTPUT_BASE_FILENAME}.wxs" -t vscodium.xsl -gg -sfrag -scom -sreg -srd -ke -cg "AppFiles" -var var.AppName -var var.ProductVersion -var var.IconDir -var var.LicenseDir -var var.BinaryDir -dr APPLICATIONFOLDER -platform "${PLATFORM}"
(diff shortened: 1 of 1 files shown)


GitMensch:
It is correct to don't have that in productversion, but the patch is not correct.

Can you please give it a second try to remove it from MS_TAG early on and - if it isn't part of that yet - add it to PRODUCT_NAME as "- Insiders" instead (and ensure that PRODUCT_ID is also different from the normal build [if that is a string then as "-insiders"]? This would match MS' insider build installer.

eternalphane:
> and ensure that PRODUCT_ID is also different from the normal build [if that is a string then as "-insiders"]?

Seems like `PRODUCT_ID` should be GUID. Changing `PRODUCT_NAME` should be enough if insider build is not to be added to CI build.

GitMensch:
We all hope for CI insider builds, so please adjust `PRODUCT_ID`, too.

GitMensch:
Friendly ping @eternalphane about fixing PRODUCT_ID along this and recheck your wix build, I'll merge it afterwards.

eternalphane:
sorry. been a bit busy IRL. will take a look this weekend.

GitMensch:
> sorry. been a bit busy IRL. will take a look this weekend.

No problem. Any update though?

daiyam:
I've integrated the fix in #1217.

PR #997, opened 2022-02-15
changed hardcoded owner

dsm:
Hi, This may be well for vscodium fork.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -27,10 +27,14 @@ do
     EXIT_STATUS=$?
     echo "exit: $EXIT_STATUS"
 
+    OWNER="${GITHUB_REPOSITORY_OWNER:-"VSCodium"}"
+    REPO_NAME=${GITHUB_REPOSITORY:(${#OWNER}+1)}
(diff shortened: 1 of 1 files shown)


GitMensch:
Where is that variable set?

dsm:
https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

This are github's default env vars and github says that;

"The default environment variables that GitHub sets are available to every step in a workflow.

We strongly recommend that actions use environment variables to access the filesystem rather than using hardcoded file paths. GitHub sets environment variables for actions to use in all runner environments."

GitMensch:
Thank you for the information. Please adjust to pre-calculate two variables - OWNER and REPOSITORY (from the env vars available the first must be removed from the second) with a fallback to the current values - and I'll merge. Thanks.

dsm:
I made this changes available please check it. Thanks

daiyam:
I've moved the variables outside the loop (#1002)

PR #1017, opened 2022-03-16
build: use windows 2019 since 2016 is deprecated

daiyam:


daiyam:
Using windows-2022, the build is failing.

PR #1029, opened 2022-03-31
fix: telemetry is not completely disabled

honbey:
The script can't find `'telemetry.enableTelemetry':`, so it not replace true with false.


stripedpajamas (approved):

daiyam:
After reviewing, since `telemetry.enableTelemetry` has been replaced by `telemetry.telemetryLevel`, the whole logics can been remove. `telemetry.telemetryLevel` is still to `off` by default.
So the PR isn't needed anymore... Thx anyway for it.

PR #1042, opened 2022-04-08
fix: letterpress icons names

GeopJr:
Upstream seems to have renamed the letterpress icons causing code-oss icon to show up on light and hc themes:

- high contrast got split into `hcLight` & `hcDark`
- `letterpress.svg` got renamed to `letterpress-light.svg`

`hcLight` is a copy of `hcDark` and the colors are taken from upstream

([microsoft/vscode@main/src/vs/workbench/browser/parts/editor/media/](https://github.com/microsoft/vscode/tree/main/src/vs/workbench/browser/parts/editor/media))


daiyam:
@GeopJr Thx

PR #1046, opened 2022-04-11
ci: bump github actions

Fdawgs:
Bump GitHub Actions to their latest major versions. Main change is they're now using Node 16 instead of 12, which is EOL at end of month.

- Changelog for `actions/checkout` can be [found here](https://github.com/actions/checkout/blob/main/CHANGELOG.md)
- Releases for `actions/setup-node` can be [found here](https://github.com/actions/setup-node/releases)
- Releases for `actions/cache` can be [found here](https://github.com/actions/cache/releases)
- Releases for `actions/setup-python` can be [found here](https://github.com/actions/setup-python/releases)

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -32,13 +32,13 @@ jobs:
           image: vscodium/vscodium-linux-build-agent:stretch-armhf
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Install GH
         run: ./install_gh.sh
 
       - name: Setup Node.js environment
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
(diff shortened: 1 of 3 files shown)


paulcarroty:
```
uses: actions/setup-node@v3
        with:
          node-version: 14
```
> they're now using Node 16 instead of 12

Node 16 or maybe v14?

daiyam:
@paulcarroty He's talking about the node version inside the github actions not the version used to build VSCodium...

Yes, at first, I was also confused since we are using node v14 and vscode has just moved to v16 (v14 is still valid but an update is needed)...

paulcarroty:
Ohh, figured out now, thx. This can be merged if CI will be OK.

daiyam:
@Fdawgs Thx

PR #1054, opened 2022-04-19
build: use node v16

daiyam:


paulcarroty:
We need to update Python too: `gyp ERR! find Python - version is 3.5.3 - should be >=3.6.0`

daiyam:
Yep, I've seen that. It's due to not up to date built agents...

PR #1108, opened 2022-06-06
build(stores/snapcraft): quote grep pattern so shell won't interpret it

Fdawgs:
Regex passed to grep frequently contains characters that collide with globs, quoting it stops that from happening.


daiyam:
Yep, the search is for the string `stable*`.

PR #1122, opened 2022-06-14
Update README.md

Dmole:
bash example for arch like all the other distros


stech11845 (changes requested):
Could you add "bash example" before the example?

Dmole:
> `yay` is only one of the available [AUR helpers](https://wiki.archlinux.org/title/AUR_helpers). I don't think we should prioritize one over another one.

On Debian there is apt-get, apt, aptitude, synaptic, etc. There should be one copy paste example per Distro.

paulcarroty:
It's more suitable for `/docs/install.md`, we don't need bloated README.

PR #1095, opened 2022-05-31
docs: Update readme for downloaded packages

xsrvmy:
Change the reference to VS Marketplace to GH repos, since the process upstream has changed.

PR #1133, opened 2022-06-18
patch: use github versions of included extensions

xaberus:
This disables the use of the marketplace to download the bundled extensions and uses GitHub links instead.

@daiyam I played around with the patch set. It feels a little fragile because there is no version dependence between the VSCodium tree and the VSCode release (like a commit reference of similar). (Also, using the update [endpoint](https://update.code.visualstudio.com/api/update/) is a potential TOS violation...)

I wrote a little script to apply and/or regenerate the patch set using (GitPython). I can clean it up and post it somewhere, if there is interest.
```
[14:03:05] Downloading extension from GH: ms-vscode.js-debug-companion@1.0.18 ...
[14:03:05] Downloading extension from GH: ms-vscode.js-debug@1.68.0 ...
[14:03:05] Downloading extension from GH: ms-vscode.vscode-js-profile-table@1.0.2 ...
```

This should fix the extension part of [#1095](https://github.com/VSCodium/vscodium/pull/1095).

### DIFF
--- a/patches/use-github-extensions.patch
+++ b/patches/use-github-extensions.patch
@@ -1,0 +1,12 @@
+diff --git a/build/gulpfile.extensions.js b/build/gulpfile.extensions.js
+--- a/build/gulpfile.extensions.js
++++ b/build/gulpfile.extensions.js
+@@ -223,7 +223,7 @@ const cleanExtensionsBuildTask = task.define('clean-extensions-build', util.rimr
+ const compileExtensionsBuildTask = task.define('compile-extensions-build', task.series(
+ 	cleanExtensionsBuildTask,
+ 	task.define('bundle-extensions-build', () => ext.packageLocalExtensionsStream(false).pipe(gulp.dest('.build'))),
+-	task.define('bundle-marketplace-extensions-build', () => ext.packageMarketplaceExtensionsStream(false, product.extensionsGallery?.serviceUrl).pipe(gulp.dest('.build'))),
++	task.define('bundle-marketplace-extensions-build', () => ext.packageMarketplaceExtensionsStream(false).pipe(gulp.dest('.build'))),
+ ));
+ 
+ gulp.task(compileExtensionsBuildTask);


daiyam:
@xaberus Could you rename the patch as `use-github-extensions.patch`? I think it's a bit clearer of the patch's purpose. Thx

daiyam:
Was my patching process helpful? What about your script?

xaberus:
The patch name somehow ended up different than I intended in the original commit. I force pushed a new version of the commit with your recommended name. 

The patch process was helpful, thank you! I have the following thoughts (maybe this is not the best place to discuss this, there was a matrix room?)
- The patches have no well defined order. As long as they are not touching the same file this is alright, but might get problematic in the future.
- The index line in the patches serves no purpose and could be removed.
- It would be nice if the patch sets base commit of the vscode tree  (`MS_COMMIT`) were saved somewhere in the repo. On the one hand, using the markeplace endpoint to determine the commit is probably not allowed by the TOS. On the other hand it, makes it harder to reproduce a build that is not current (I am thinking for example about git bisect).

My script right now uses GitPython to
- Create a shallow clone of the vscode tree using the `MS_COMMIT` provided (similar how the fetch part of `get_repo.sh` does it) and then applies the patches to it in (random) order. Like I said above, patch ordering would be nice here (`0001-do-stuff.patch`).
- Then I can fetch some other commit and use git rebase to rebase the patch set using the full toolkit git provides.
- Then I can use the script to export the whole patch set again.

I hope this makes sense. I am cleaning it up right now so it will work with parameters from the command line instead of hard-coded constants.

daiyam:
> The patches have no well defined order. As long as they are not touching the same file this is alright, but might get problematic in the future.

The patches are executed in alphabetical order. If there is an issue with a future patch, we will look at the possible solutions at that time.

> The index line in the patches serves no purpose and could be removed.

It is generated by the command `git diff` and doesn't cause any trouble. So we can leave it there.

> It would be nice if the patch sets base commit of the vscode tree (MS_COMMIT) were saved somewhere in the repo.

I agree but there was no demand for building an older version.
In any case, you could use the tag of respective version to build it.


Anyway, feel free to make any PR for things you would like to improve.

GitMensch:
> The patches are executed in alphabetical order. If there is an issue with a future patch, we will look at the possible solutions at that time.

... in which case the solution is to apply what MSYS2 and other package manager do: prefix the patches with a number and an underscore.

Issue #1165, opened 2022-07-21
Script undo_telemtry.sh takes a long time, parsing though the entire node_modules tree


cedricroijakkers:
This bug has been reported by users of the vscodium AUR package: https://aur.archlinux.org/packages/vscodium

When building vscodium, the script `undo_telemtry.sh` runs as part of the build script.

However, it seems this does a recursive grep through the entire `node_modules` directory, which is relatively large, and contains a lot of files. This grep takes quite a long time (depending on the building machine of course, but even on my relatively new and powerful Dell laptop it takes over half an hour). During that times it consumes 1 full CPU.

Is there no way to make this grep smarter and/or faster? Grepping through all files in the `node_modules` seems a bit overkill.

- [X] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

- [X] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**To Reproduce**
Build vscodium. The build will work, but it will take a long time. During the `undo_telemtry.sh` step, check the running processes, and you will find a grep consuming 100% CPU.

**Expected behavior**
The grep should not take such a long time.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Architecture x64
 - Version 1.69.1 and 1.69.2

**Additional context**
N/A

MithicSpirit:
For me it takes a lot longer to run the subsequent `sed` command than `grep` did. This could probably be sped up by telling `xargs` to run multiple `sed` processes in parallel, though it would also be important to have some way to limit the number of simultaneous jobs this spawns (which is possible in `xargs` iirc).

cedricroijakkers:
@MithicSpirit You are correct actually, it's the `sed` that is eating up the most CPU/time, not the grep.

Still, I'm not sure if just throwing more processes at it is the solution here, seems like there is more to gain by filtering the type of files that `sed` should work on, instead of just making its way through the entire `node_modules`?

MithicSpirit:
Considering that it's only replacing Microsoft telemetry URLs and that the majority of the time is spent in `sed`, which is only run on files that have already been confirmed to have a Microsoft telemetry URL, I don't think that the problem is that there are too many files being unnecessarily considered. If that were the case, the main time sink would be `grep`.

rmatab:
It has taken over one hour already, and it is still running with no way of knowing how much more time it will take:

[...]

PR #1170, opened 2022-07-22
fix: replace `grep` with `ripgrep` due to excruciating timing

daiyam:
This PR is replacing `grep` with `ripgrep` due to very slow executing timing...

Fixes: #1165


HTC37 (commented):
gh pr checkout 1170

daiyam:
Here my issue for armv7:
```
ELF Header:
  Magic:   7f 45 4c 46 01 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF32
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           ARM
  Version:                           0x1
  Entry point address:               0x28c31
  Start of program headers:          52 (bytes into file)
  Start of section headers:          4060792 (bytes into file)
  Flags:                             0x5000400, Version5 EABI, hard-float ABI
  Size of this header:               52 (bytes)
  Size of program headers:           32 (bytes)
  Number of program headers:         10
  Size of section headers:           40 (bytes)
  Number of section headers:         31
  Section header string table index: 30
../undo_telemetry.sh: line 37: /usr/bin/rg: cannot execute binary file: Exec format error
```

PR #1217, opened 2022-08-24
feat: insider version

daiyam:
This PR is preparing for the insider version with few improvements:
- separated build scripts (per platform) are regrouped as one
- simplified `prepare_vscode.sh` for updating the `product.json` properties

The keep tracks of which commit/version the patches have been built against, the `insider.json` will be used.

PR #1226, opened 2022-08-31
Fix README logo

DvvCz:
Noticed the logo was a dead link on my profile, so I fixed it pointing to the wrong path since PR #1217


daiyam:
Yep, I've seen it yesterday. I didn't had the courage to make the change...

PR #1251, opened 2022-09-15
docs: clarify winget install command

osfanbuff63:
`winget` uses IDs and `winget install vscodium` *might* work, but usually it's better to use `winget install -e --id VSCodium.VSCodium` to make sure it will install the correct package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,8 @@ brew install --cask vscodium
 #### <a id="install-with-winget"></a>Install with Windows Package Manager (WinGet)
 
 If you use Windows and have [Windows Package Manager](https://github.com/microsoft/winget-cli) installed:
-```bash
-winget install vscodium
+```cmd
+winget install -e --id VSCodium.VSCodium
 ```
 
 #### <a id="install-with-choco"></a>Install with Chocolatey (Windows)


daiyam:
@GitMensch @vedantmgoyal2009 What do you think? I think it's fine as it is.

vedantmgoyal2009:
As long as any other contributors doesn't play with the manifests and change the `Moniker` field, it is fine as it is.

AlexisComix:
From what I have tested, `winget install vscodium` installs from VSCodium.VSCodium fine. Changing ` ```bash` to ` ```cmd` makes no functional difference because `cmd` is not recognised by Linguist. However, changing it to something other than `bash` could be good as Windows does not use bash without WSL. While a change to ` ```bat` or ` ```batch` would also have no visual difference to ` ```bash`, it is recognised by Linguist and would make sense as ` ```bash` may be slightly misleading.

GitMensch:
> Changing `` ```bash `` to `` ```cmd `` makes no functional difference because `cmd` is not recognised by Linguist. However, changing it to something other than `bash` could be good as Windows does not use bash without WSL. While a change to `` ```bat `` or `` ```batch `` would also have no visual difference to `` ```bash ``, it is recognised by Linguist and would make sense as `` ```bash `` may be slightly misleading.

I _think_ you've meant the install docs and the code fence used there - I've just changed all three windows entries from "bash" to "cmd".

PR #1274, opened 2022-09-30
feat: show announcements to welcome page

daiyam:
This PR shows a list of announcements on welcome page with items found in https://github.com/VSCodium/vscodium/blob/insider/annoucements.json (depend on the quality)

@GitMensch @paulcarroty What do you think?

### DIFF
--- a/patches/feat-announcements.patch
+++ b/patches/feat-announcements.patch
@@ -4,32 +4,32 @@ index 9564618..35cde81 100644
 +++ b/src/vs/workbench/contrib/welcomeGettingStarted/browser/gettingStarted.ts
 @@ -116,2 +116,3 @@ type GettingStartedActionEvent = {
  type RecentEntry = (IRecentFolder | IRecentWorkspace) & { id: string };
-+type AnnoucementEntry = { id: string, title: string, url: string };
++type AnnouncementEntry = { id: string, title: string, url: string };
  
 @@ -148,2 +149,4 @@ export class GettingStartedPage extends EditorPane {
  	private gettingStartedList?: GettingStartedIndexList<IResolvedWalkthrough>;
-+	private annoucementList?: GettingStartedIndexList<AnnoucementEntry>;
-+	private annoucementData?: AnnoucementEntry[];
++	private announcementList?: GettingStartedIndexList<AnnouncementEntry>;
++	private announcementData?: AnnouncementEntry[];
  
 @@ -767,2 +770,3 @@ export class GettingStartedPage extends EditorPane {
  		const gettingStartedList = this.buildGettingStartedWalkthroughsList();
-+		const annoucementList = await this.buildAnnoucementList();
++		const announcementList = await this.buildAnnouncementList();
  
 @@ -777,3 +781,3 @@ export class GettingStartedPage extends EditorPane {
  				this.container.classList.remove('noWalkthroughs');
 -				reset(leftColumn, startList.getDomElement(), recentList.getDomElement());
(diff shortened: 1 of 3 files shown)


patches/feat-announcements.patch
```diff
@@ -0,0 +1,69 @@
...
++			return li;
++		};
++
++		if (this.announcementList) { this.announcementList.dispose(); }
++
++		const announcementList = this.announcementList = new GettingStartedIndexList({
++			title: localize('announcements', "VSCodium Announcements"),
++			klass: 'announcements',
++			limit: 5,
++			empty: $('.empty-recent', {}, localize('noAnnouncements', "There is no announcements.")),
```
GitMensch, line 50: Should be "There is no current announcement."

patches/feat-announcements.patch
```diff
@@ -0,0 +1,69 @@
...
++			const anchor: HTMLLinkElement = $('a');
++			anchor.href = url;
++			anchor.innerText = title;
++			anchor.target = '_blank';
++			li.appendChild(anchor);
++
++			return li;
++		};
++
++		if (this.announcementList) { this.announcementList.dispose(); }
```
GitMensch, line 44: If we dispose the `announcementList, then `announcementData` should be disposed, too (or otherwise both kept)

GitMensch:
I like the general idea, but I'm not sure about how the data gets in; Wouldn't it be reasonable to at least have "static" data (that refers to the installed release), we could package a file during install and read that locally and _maybe additional_ have dynamic data?
And even with dynamic data some users will not want to have automatic file downloads (and some work without internet access in any case); so maybe cache that and get it only every N days at start [N may be 1 successful, too]?

In any case it would be good to have a screenshot how that looks like using the current announcement.json (and maybe squashing the commits before the merge).

patches/feat-announcements.patch
daiyam: In my test, the `buildAnnouncementList()` function was called 3 times to display the page... that why I'm caching the data. So it's fine like that.

patches/feat-announcements.patch
daiyam: In my test, the buildAnnouncementList() function was called 3 times to display the page... that why I'm caching the data. So it's fine like that.

daiyam:
> Wouldn't it be reasonable to at least have "static" data (that refers to the installed release), we could package a file during install and read that locally and maybe additional have dynamic data?

Ya, but a some dynamic data is a must so we can inform the users of possible issues. So maybe a mixed approach.

> some users will not want to have automatic file downloads

I thought of that, so maybe with a new setting??

> and some work without internet access in any case

The list will be empty

> so maybe cache that and get it only every N days at start [N may be 1 successful, too]?

I think it's unnecessary.

> In any case it would be good to have a screenshot how that looks like using the current announcement.json
Screen Shot 2022-10-01 at 1 25 14 AM

> and maybe squashing the commits before the merge

Yes, I always do that when I can.

patches/feat-announcements.patch
paulcarroty: "There are ..."

paulcarroty:
An option to disable dynamic data would be great.

patches/feat-announcements.patch
daiyam: Yep, I will fix my grammar mistake.

daiyam:
I've added the`workbench.welcomePage.extraAnnouncements` setting to enable/disable the "dynamic" data.
The `announcements-builtin.json` file contains the announcements that are included by default.
The `announcements-extra.json` file will be fetched when the page is shown.

GitMensch (approved):
The new one looks very good to me, Thank you very much!

Issue #1281, opened 2022-10-07
Wrong information shown into "About" dialog after install russian language pack


sergey-dryabzhinsky:
**Describe the bug**
After VSC moved to release+commit verioning lines in About dialog shifted and now shown wrong information.

**Please confirm that this problem is VSCodium-specific**
- [x] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

**Please confirm that the issue/resolution isn't already documented**
- [x] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**To Reproduce**
0. Setup language pack. In my case - russian.
1. Go to main menu 'Help/Помощь'
2. Click on 'About/О программе'
3. See wrong info in dialog

**Expected behavior**
Build date in Date info, etc.
Seems like russian lang pack does not have "Release:" line. And all other lines shifted.

**Screenshots**
![Снимок экрана_2022-10-07_20-52-11](https://user-images.githubusercontent.com/717597/194621873-1dc11dc8-5f2f-43bf-aacf-9a30ed6462aa.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu Linux 18.04
 - Architecture: amd64
 - Version: since 1.70.1+
 - App Manager: tar.gz

**Additional context**
If I remove language pack - all info got in right lines.
![Снимок экрана_2022-10-07_21-06-32](https://user-images.githubusercontent.com/717597/194621848-83c3f486-54ab-4c3d-83cd-5609b6260c5d.png)

daiyam:
@sergey-dryabzhinsky This issue is specific to VSCodium since the release number isn't available on VSCode.

Crystal-RainSlide:
Also see:

https://github.com/VSCodium/vscodium/issues/1301

https://github.com/VSCodium/vscodium/blob/10582b3c86c800f43dbe7836203420dde9862c60/patches/build-version.patch#L145

https://github.com/microsoft/vscode/blob/c07b428d7db4662a1f45e8e57e41d44167f2e11a/src/vs/workbench/electron-sandbox/parts/dialogs/dialogHandler.ts#L169

daiyam:
@Crystal-RainSlide Feel free to make a PR to fix the issue 😉

Crystal-RainSlide:
> Honestly, I don't know...
> - Should we hide the Release version?
> - Should we merge the Release version into the "Version:"?
[...]

PR #1305, opened 2022-10-26
Fix: add translation for release (#1281)

Crystal-RainSlide:
This fixes https://github.com/VSCodium/vscodium/issues/1281, however translation to `Release` (or `Release version`, if the word `Release` in that lang can't represent a version number) for other language pack needs to be added.

In the future, If there are more strings requires translation, a refactor is needed to concentrate them and apply them at once.

### DIFF
--- a/patch.sh
+++ b/patch.sh
@@ -13,7 +13,8 @@ fi
 
 read -p "Press any key when the conflict have been resolved..." -n1 -s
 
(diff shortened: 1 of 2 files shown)


Crystal-RainSlide:
- Need test to see if this works;
- Maybe wait a while to gather more translations.

PR #1344, opened 2022-12-02
feat: add default apis for open-remote-ssh

daiyam:
This PR is adding default apis for [open-remote-ssh](https://open-vsx.org/extension/jeanp413/open-remote-ssh) so it works out of the box.

https://github.com/jeanp413/open-remote-ssh/issues/37

### DIFF
--- a/build/update_api.sh
+++ b/build/update_api.sh
@@ -30,4 +30,6 @@ fi
 
 APIS=`cat ${DIRECTORY}/resources/app/product.json | jq -r '.extensionEnabledApiProposals'`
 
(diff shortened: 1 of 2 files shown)


GitMensch (approved):
LGTM (haven't checked, but assume that the list _is_ identical to `ms-vscode-remote.remote-ssh`)

GitMensch:
Just checking: Using the win x86_64 .zip distribution....

When removing the previously added

```jsonc
     "enable-proposed-api": [
         "jeanp413.open-remote-ssh",
     ]
```

from my argv.json, then open-remote-ssh did not get activated; should this work?

GitMensch:
Don't mind, works like a charm, somehow I've opened the too old vscodium. When using 1.74.1 the extension opens without that setting.

PR #1370, opened 2023-01-07
fix: patch mangle for windows

daiyam:
This PR adds a patch to fix a mixup when mangling typescript files on Windows.
`esbuild` generates paths with `\` separator while the mangler uses `/`, so the content can't be updated since the paths aren't matching

Fixes https://github.com/jeanp413/open-remote-ssh/issues/44

VSCode doesn't suffer from this issue since the typescript files are compiled once (on Ubuntu) and then shared when the different versions are built for each platform...

### DIFF
--- a/patches/fix-mangle-windows.patch
+++ b/patches/fix-mangle-windows.patch
@@ -1,0 +1,9 @@
+diff --git a/build/lib/compilation.ts b/build/lib/compilation.ts
+index 2cc04d2c977..9029c6231ed 100644
+--- a/build/lib/compilation.ts
++++ b/build/lib/compilation.ts
+@@ -128,3 +128,3 @@ export function compileTask(src: string, out: string, build: boolean): () => Nod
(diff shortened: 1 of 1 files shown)


xaberus:
In your PR you changed just the `compilation.ts`. Is this enough? I.e., will the corresponding `compilation.js` file be generated somewhere else automatically? Are they commiting the generated ``compilation.js`` file  just in case?

daiyam:
You are right! I hate that!

xaberus:
OK, so it is not just me 🤣.

daiyam:
done with https://github.com/VSCodium/vscodium/commit/57507e0abfd94e29355487af52043f7f1ed9e3e3

daiyam:
@xaberus  Thx for the catch

daiyam:
JFYI, this patch have created a regression of that we aren't able to build the Insiders version on Windows...

PR #1426, opened 2023-02-26
Create dependabot.yml

Araxeus:
I was looking through the GitHub actions and noticed that you are using out of date actions, for example:

https://github.com/VSCodium/vscodium/blob/b405f160de548f07534301a00de9dbc23046c582/.github/workflows/stable-windows.yml#L124

This dependabot configuration will open PR's to bump those action versions

> I thought about just bumping `winget-releaser` to v2 but this a better long term solution IMO

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


GitMensch:
Maybe check less often than `daily`?

GitMensch:
Note: the build workflow (will be un-broken by #1432) is not needed for this PR.

Araxeus:
> Maybe check less often than `daily`?

How often do you want it to run?

daiyam:
> How often do you want it to run?

1st, 15th on the month should be enough

Araxeus:
> > How often do you want it to run?
> 
> 1st, 15th on the month should be enough

That's a very low frequency, any reason not do it weekly?

GitMensch:
The main reason would be not to create PRs too often as this keeps the maintainers busy. But as this is about GH actions only and I _think_ that there won't be too much deprecation so often and dependabot won't take too much CI minutes in any case I'm fine with a weekly cycle.

Araxeus:
Like you said this only checks github actions, so PR's would be very rare anyways ;)

I'll change it to weekly soon

PR #1432, opened 2023-03-02
fix: update to 1.76

daiyam:


GitMensch:
It will be interesting to see if OpenSSH-Remote still works... ;-)

daiyam:
I will test it on the Insiders. But anyway, the `1.76` will be pushed tomorrow. If the workflows run now, they might create 2 releases... I don't want to try my luck.

daiyam:
I wasn't to compile the project on the Mac mini M1 (which I use to update VSCodium since it's fast). Fixed now.

daiyam:
Building, going to test it in the morning.

PR #1457, opened 2023-04-01
Fix MacOS Icon generation

### LINKED ISSUE #1456: New icon doesn't look right on MacOS
**Describe the bug**
The new icon seems not to be centered on MacOS. It seems to be cropped out on the bottom and on the sides.

- [X] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

It's specific to the new icon, so yes.

- [X] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/DOCS.md) and my issue is not mentioned there.

**To Reproduce**
- If you have VSCodium installed, just install the last update (which includes the new icons) and relaunch VSCodium.
- If not, just download it and open the .dmg file.

**Screenshots**
Screenshot 2023-04-01 at 13 15 12

For comparison (Screenshot took from the [icon discussion](https://github.com/VSCodium/vscodium/discussions/1429)): 
Screenshot 2023-04-01 at 13 19 41

 - OS: MacOS
 - Version 1.77.0

**App managers tested**
Homebrew, also tested without app managers.

NovaliX-Dev:
Fix #1456.

With this fix, i've done other things: 
- Made the logo a little smaller because it felt too big in my opinion. I used `-gravity center` instead of `-geometry +<x>+<y>` option to center it.
- added the dependencies needed for icon generation in the build docs, since there weren't there.

## Comparison
Before :

After :

**Note** : For some reasons the `png2icns` program didn't work on my machine, so i used the generated pngs instead for the comparison. If you can run it, please do it to see if there is no other issues.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -38,6 +38,10 @@
 - imagemagick (for AppImage)
 
 ### <a id="dependencies-macos"></a>MacOS
+- imagemagick
+- png2icns (`npm install png2icns -g`)
+- librsvg
+
 
 ### <a id="dependencies-windows"></a>Windows
 
--- a/icons/build_icons.sh
+++ b/icons/build_icons.sh
@@ -33,8 +33,8 @@ VSCODE_PREFIX=""
 
 build_darwin_main() { # {{{
   if [ ! -f "${SRC_PREFIX}src/${QUALITY}/resources/darwin/code.icns" ]; then
-    rsvg-convert -w 700 -h 700 "icons/${QUALITY}/codium.svg" -o "code_logo.png"
-    composite "code_logo.png" -geometry +165+190 "icons/template_macos.png" "code_1024.png"
+    rsvg-convert -w 655 -h 655 "icons/${QUALITY}/codium.svg" -o "code_logo.png"
+    composite "code_logo.png" -gravity center "icons/template_macos.png" "code_1024.png"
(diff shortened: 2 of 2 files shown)


daiyam:
Screen Shot 2023-04-01 at 3 16 23 PM

Here the reason why it isn't centered.

But a mistake was made... it should have been `155` and not `165`...

NovaliX-Dev:
I don't really understand with the image... can you explain ?

daiyam:
I might be wrong! Not a designer here...

I tried to constraint the icon the circle.
But after re-reading https://developer.apple.com/design/human-interface-guidelines/foundations/app-icons/, it seems just used for the size of the icon... 

So maybe centered with a width of 720px.

NovaliX-Dev:
Here is what it looks like with a width of 720px.

code_512

I think the logo is taking too much space here. What do you think about it ?

daiyam:
I don't know. I've checked VSCode icon: 720px.
The initial icon doesn't shock me (maybe due to my big screen and small dock).

I would like to wait for more feedback to know what to do.

Anyway, thank you for the PR!

NovaliX-Dev:
Maybe asking people like for the shapes should do it. What do you think ?

DanielProg39:
Well, I think that a new proposed icon (with smaller and more centred logo) looks better

Issue #1264, opened 2022-09-19
Support for Win7/8?


LexaGV:
I use Windows7 x64 and VSCodium WAS working on this platform. Why it doesn't work anymore? Is it possible to build Electron for Win7? Obviously there is NOTHING what Electron needs from Win10 what doesn't exist on Win7.

If it's not possible to recompile Electron, it could be nice AT LEAST keep mention of latest release which worked on Win7. Please, update README.

africalimedrop:
See https://github.com/VSCodium/vscodium/issues/1256
They need to do more than 'update the readme', they need to not allow the update on Win 7. We don't need our environments seamlessly fucked.

paulcarroty:
Waiting for https://github.com/electron/electron/issues/35597.

daiyam:
I tried to find a doc for the system requirements to run electron (not to build it). I didn't found it...

zedayseekay:
> 

Windows 7 support is mentioned in https://github.com/electron/electron/blob/main/README.md

Under https://github.com/electron/electron/issues/35597 I have now added a question on whether there is a more formal support matrix - I couldn't find one either, you're right! It doesn't matter in this case because https://github.com/electron/electron/issues/35597 has been treated as valid anyway.

LexaGV:
It's even better news!! If Electron can (and must) run on Win7, then Codium has no problem at all to work on Win7 too! Am I right?

GitMensch:
Note that ["Visual Studio Code" 1.71+ _explicit_ does not work with Windows 7 any more](https://code.visualstudio.com/docs/supporting/faq#_can-i-run-vs-code-on-windows-7), because
> Microsoft [ended support for Windows 7](https://docs.microsoft.com/lifecycle/products/windows-7) in January, 2020 and no longer provides security updates.

If you need an installation now it is likely best to use https://github.com/VSCodium/vscodium/releases/tag/1.70.2.22230 while the upstream Electron issue is solved - and ignore any updates until this issue is closed and a follow-up release is finished.

musadarj:
So the latest working version for Win 7 (for now) is [1.70.2.22230](https://github.com/VSCodium/vscodium/releases/tag/1.70.2.22230), well I'll note that for future reference.

LexaGV:
Guys, thanks for answer. Will Win7 info be reflected in README please? My English is far from good to do it. :)

[...]

PR #1463, opened 2023-04-04
fix: banner uses same flag as the notification

daiyam:
This PR fixes the issue mentioned here: https://github.com/VSCodium/vscodium/issues/1264#issuecomment-1494342450


GitMensch:
Nice and clean - would this also apply upstream?

daiyam:
Yes, I will try it first on the Insiders.

PR #1530, opened 2023-05-27
fix(patches): enable electron sandbox

### LINKED ISSUE #1482: Electron sandboxing prevents basic functionality.
**Describe the bug**

Enabling Electron Sandboxing causes VSCodium to become unresponsive on reboot.

**To Reproduce**
1. Enable `Window > Experimental > Use Sandbox`
2. Restart VSCodium

**Expected behavior**

Electron sandboxing should not affect basic application functionality (e.g., displaying text on the canvas).

 - OS: Mac OS
 - Architecture arm64
 - Version 1.77.3
 - App Manager Homebrew

rxri:
The electron sandbox disables the node integration within the renderer process. However, the [`custom-gallery`](https://github.com/VSCodium/vscodium/blob/master/patches/custom-gallery.patch) patch required the use of the `require()` function, causing the renderer to fail loading within the sandbox. To resolve this issue, the function that requires user's `product.json` has been moved to [`main.js`](https://github.com/microsoft/vscode/blob/main/src/main.js), and the user's `product.json` is now assigned to global scope which can be used in [`product.ts`](https://github.com/microsoft/vscode/blob/main/src/vs/platform/product/common/product.ts).

Additionally, I noticed `git apply` occasionally failing to apply the patches due to blank lines on Windows. As a solution it has been changed to GNU patch command

Fixes #1482
This patch works on stable (1.78.2) and insider (1.79.0) versions


daiyam:
@rxri Thanks a lot

daiyam:
The issue with `git apply` was due to `merge-user-product.patch` being a bit corrupted (the space were missing, mostly due to the setting `Files: Trim Trailing Whitespace` enabled).
To avoid that issue, in [.editorconfig](https://github.com/VSCodium/vscodium/blob/master/.editorconfig), the behaviour has been disabled for those files. ([EditorConfig for VS Code](https://open-vsx.org/extension/EditorConfig/EditorConfig) needs to be installed)

rxri:
The git apply issue wasn't related to this patch but any other that had blank lines. It only happened on windows tho so I don't know what was the issue but I found out that it's git bug

PR #1555, opened 2023-06-26
docs: add slashes to shell script calls

ammernico:


daiyam:
`. get_repo.sh` is correct.

https://how.wtf/executing-a-script-with-dot-vs-source-in-bash.html

PR #1559, opened 2023-07-04
Optimizations for shell scripts

zpiatt:
1. Clarify return code parsing. 
    `if ! git apply --ignore-whitespace "${FILE}"; then` 
instead of 
    `git apply --ignore-whitespace "${FILE}";
    if [ $? -ne 0 ]; then`
2. Standardize shebang. Currently all `.sh` files have either `#!/bin/bash` or `/usr/bin/env bash`. Practically this has no effect, however if I'm there, I'm going to change it the preferred `#!/usr/bin/env bash`.
3. Standardize tests. Currently most `.sh` files have a mixture of `[ ]` and `[[ ]]` tests. The bash tests are potentially faster, and since it doesn't appear POSIX compatibility is a requirement I've updated them.

This is the tip of the iceberg for these types of optimizations, but I want to gauge interest in this type of contribution.

### DIFF
--- a/build/update_api.sh
+++ b/build/update_api.sh
@@ -1,23 +1,25 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
 export VSCODE_QUALITY="stable"
 
-while getopts ":ilp" opt; do
+while getopts ":i" opt; do
   case "$opt" in
     i)
       export VSCODE_QUALITY="insider"
+      ;;
+    *)
       ;;
   esac
 done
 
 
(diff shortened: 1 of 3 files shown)


zpiatt:
I fixed a potential error in `build/update_api.sh`. Line 35 had been read/writing from the same pipeline which could unexpectedly truncate the file (https://www.shellcheck.net/wiki/SC2094).

There were a lot of superfluous pipes in this script and I removed as many as practical. Notably, line 31 was using `cat <file> | jq` instead of simply `jq <file>`.

I also removed echoed variables piped into `sed` in favor of parameter expansions.

Finally, I removed options from `getopts` that weren't being parsed, and added an empty wildcard. Perhaps options were meant to be added, in the meantime, this is more explicit.

PR #1575, opened 2023-07-23
fix: bump snap base to core22

abitrolly:
I don't know if that still builds. Just want to remove `core18` dependency from my system.


daiyam:
Visual Studio Code is using `core20` since March.
We could make the move but the build script and the launcher will need to be rewritten.

abitrolly:
Could https://snapcraft.io/docs/electron-apps be helpful here?

daiyam:
`core20` is the target. Closing in favor of #1658.

PR #1582, opened 2023-07-28
fixup: remove mangling in extensions/ (introduced in 1.80)

brjsp:
This also fixes the reproducibility issue mentioned in https://github.com/microsoft/vscode/issues/189124


daiyam:
Is this still an issue after their fix?

brjsp:
Probably isn't. I haven't checked but AFAIK you want to remove mangling anyway to make hacking the UI possible.

I myself don't use UI-changing extensions but have noticed that the mangler often OOMed on our (openSUSE) servers leading to build failures.

daiyam:
The changes have been merged in #2090. We will test it in the Insiders version to see if there is no issues before moving it to the stable version.

daiyam:
@brjsp Thanks for the PR.
Sorry, seeing you have reacted, I could have asked you to make the change. But since it's a more than a year old PR, I didn't think you were still there ;)

PR #1598, opened 2023-08-11
Reduce Global Variable Assignment

zpiatt:
I'm working on reducing global variable assignment which should reduce the overhead on your build process. A nice side-effect in this particular instance was removing some repetitive code. If you like this, I'll keep plugging away at it as I have time.

1. Created `apply_patches` function. This allows for assigning `file` locally as it's not needed after the patches are applied.
    - Note: in this case I left `${1}` unquoted so the loop will iterate over the glob ([https://www.shellcheck.net/wiki/SC2066](https://www.shellcheck.net/wiki/SC2066)).
    - Also note, I double quoted the file paths. We know there are no spaces in this file path, but it's still a good habit in case there is.
2. I removed the `setpath_json` function. The only difference in the `setpath` functions is `--arg` vs `--argjson`. I made `--argson` an argument in the applicable lines and used `--arg` as the default value. This makes no difference outside of removing repeated code, so I'm happy to revert if you don't like it.

### DIFF
--- a/icons/build_icons.sh
+++ b/icons/build_icons.sh
@@ -21,6 +21,7 @@ while getopts ":i" opt; do
 done
 
 check_programs() { # {{{
+  local arg
   for arg in "$@"; do
     if ! command -v "${arg}" &> /dev/null; then
       echo "${arg} could not be found"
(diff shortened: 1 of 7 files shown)


zpiatt:
There's a potential issue with the unquoted `${1}` that will be an easy fix. I also found a few more unneeded global variables, so I'll do a PR for them at once.

PR #1608, opened 2023-08-18
[vscodium][linux] Fix the patch file for linux builds

theskcd:
The URL was updated on the vscode repo, but not updated for the patch. Fixing it on this commit.

The code which changed is here: https://github.com/microsoft/vscode/blob/main/build/linux/dependencies-generator.js#L24C1-L25C1


daiyam:
FYI, it should have been merged in the `insider` branch.

PR #1615, opened 2023-08-24
Reduce Global Variable Assignment 01

zpiatt:
I got a lot more work done on reducing global variables. I made some assumptions for variable/function names that will need to be reviewed.

1. Removed all system calls to `basename` opting for parameter expansions instead.
2. Localized variables that were already in functions, but were not required globally.
3. Created new functions to facilitate localizing variables.
    - The `apply_patches` function now iterates across the patches explicitly using `find` rather than relying on globbing. This also allows the test `[[ -f ]]` to be removed and prevents any errors white-space could introduce. 
4. Moved variables that were defined outside of a function, inside the function that was calling them.
5. Attempted to clarify when variables were meant to be global/local.
    - This is inconsistent, but it appears the intent was to make global variables all caps. Where applicable I made local variables lowercase and/or camel-case per previous style guidance. If this style was consistent it would make it a lot easier to further reduce global variables and optimize the build process further.

As always, I'm happy to make any tweaks you require.

### DIFF
--- a/icons/build_icons.sh
+++ b/icons/build_icons.sh
@@ -21,6 +21,7 @@ while getopts ":i" opt; do
 done
 
 check_programs() { # {{{
+  local arg
   for arg in "$@"; do
     if ! command -v "${arg}" &> /dev/null; then
       echo "${arg} could not be found"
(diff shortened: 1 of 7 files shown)


zpiatt:
I see there's an error with the `find` command in `apply_patches`. I mistakenly added a trailing `/`. It also occurs to me `-maxdepth` should be set so `find` doesn't enter the `insider` directory unless the conditional is passed.

zpiatt:
Is it possible there's an issue with `/patches/fix-build-linux.patch`. I'm not seeing an obvious reason this should be failing.

zpiatt:
It looks like `Run ./build.sh` has been failing the last few days. At various patches it fails as below:
```
failed to apply patch ../patches/fix-build-linux.patch
Error: Process completed with exit code 1.
```

daiyam:
@zpiatt It's due to #1608...

zpiatt:
> @zpiatt It's due to #1608...

That's what I thought. It also failed at the same step (though on a different patch) before that was merged: https://github.com/VSCodium/vscodium/actions/runs/5948672438/job/16132919035#step:9:76

Issue #1657, opened 2023-09-23
Better workflow for the Snap package


soumyaDghosh:
The current workflow for building the snap package is very complicated. It can be replaced by a very simple method. There can be a separate repo for the `snapcraft.yaml`, that'll served in the store. And to test if it the insiders build, there will be a manifest in this repo. The manifest can also be simplified, by connecting it to the `gnome-42-2204` plug.

PR #1658, opened 2023-09-23
`Snap` Ported to core22

soumyaDghosh:
Also, kindly check #1657. The workflow can be improved and armhf and other architectures can also be supported in that way. If you want, I can elaborate on how the workflow can be managed.

### DIFF
--- a/stores/snapcraft/stable/snap/snapcraft.yaml
+++ b/stores/snapcraft/stable/snap/snapcraft.yaml
@@ -4,7 +4,7 @@ summary: Code editing. Redefined.
 description: |
   Binary releases of Code without branding/telemetry/licensing
 
-base: core18
+base: core22
 grade: stable
 confinement: classic
 compression: lzo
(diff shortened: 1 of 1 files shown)


daiyam:
@soumyaDghosh Can you use `core20` instead of `core22` since VSCode is also using `core20`

soumyaDghosh:
I am rethinking on this PR. This probably needs some more changes. Classic snaps are big thing to test.

daiyam:
Following the discussion #1748 (I've looked on how flatpacking is done) and some about the AUR, I'm wondering several thinks...

- Is it possible to a channel for each version `core18`, `core20`, `core22`?
- What about a channel for X11 or Wayland
- Or which Electron to use? system or builtin? And firstly, is it possible to reference a specific version of Electron installed system wide?

Is it done by any app? Should we do it (12 versions...)?

guyluz11:
It will be very cool to support armhf and to run it on a Linux phone XD

soumyaDghosh:
Okay, @daiyam The initial porting is done. Now needs some testing before merging this. We can have some discussions on how to publish these snaps on store. We definitely don't need two different snaps for stable and latest builds. Also, this is from my last commit.

>1. remove content snap(can't be used with classic snaps)
>2. patch the rpath of the elf files to allow it run properly without
   depending on host libraries
>3. readded the apt packages and added a part to remove duplicate files
   that are in the core snap.
>
>I am still unsure if all these packages are necessary.

soumyaDghosh:
I will be using my local build as my daily driver from now, on my `Ubuntu 23.10`. If some other OS guys can also have some testing, we can merge this.

daiyam:
Does it work on Ubuntu 20?

soumyaDghosh:
@daiyam Thanks for recalling this. As suspected, it didn't. It needs more investigation.

guyluz11:
> Also, kindly check #1657. The workflow can be improved and armhf 

I've just opened the issue supports snap arm.
The comments you added was for the pr with core22.

Is it also relevant for core20?

soumyaDghosh:
Yup. It's same for all.

daiyam:
@soumyaDghosh I'm migrating to `core20`, looking at your code you have lot of changes...
I'm wondering what you are trying to fix with the custom `LD_LIBRARY_PATH`, can you tell me? Thx

soumyaDghosh:
@daiyam That's a try to pick gtk related files from the gnome-42-2204 snap.. but it's not supported for classic snaps. So, I have been looking into alternative ways. I actually have a core22 port, which I am using now.. it ran fine on 20.04.. but in my latest 23.10, I get this error message when I run it.. so, I didn't push it.

```bash
/snap/core22/current/usr/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /usr/lib/x86_64-linux-gnu/gio/modules/libgvfsdbus.so)
Failed to load module: /usr/lib/x86_64-linux-gnu/gio/modules/libgvfsdbus.so
```

daiyam:
I can't figured out my issue. I'm able to build on local (with the latest `insider` branch) but CI don't want to do it...
I would be great if you can take a look! (I'm not even on Linux...)
Also, the size of the snap is quite big (currently 380Mo but with, I think, vscodium's files in double... 300Mo without). Is it normal?

soumyaDghosh:
@daiyam Sorry, was off due to exams. So, should I push the changes I have with core22? The only issue is what I shared with you. But can you guess, why is libgvfsdbus.so file getting called there? If we can solve that, I hope that we can fix the whole issue

daiyam:
@soumyaDghosh No, we just moved to `core20`. We stay on it until MS move to `core22`.

Have you tried the latest version?

soumyaDghosh:
> Have you tried the latest version?

If you mean with `core20` probably not. But if you meant the latest release with my changes, then yes. I did.

daiyam:
> But if you meant the latest release with my changes, then yes. I did.

Yes, it was what I meant. Do you have any issue with it?

PR #1687, opened 2023-10-09
Build appimage

### LINKED ISSUE #1654: No AppImage since 1.82.1
**Describe the bug**
For 1.82.0, there is an AppImage download. For 1.82.1 and 1.82.2, this download is missing

**To Reproduce**
1. Go to https://github.com/VSCodium/vscodium/releases/tag/1.82.2.23257
2. Search for an AppImage download -> there is no such download

**Expected behavior**
1. Go to https://github.com/VSCodium/vscodium/releases/tag/1.82.0.23250
2. Search for an AppImage download -> you will find it

uli-heller:
Probably fixes #1654

### DIFF
--- a/build/linux/appimage/build.sh
+++ b/build/linux/appimage/build.sh
@@ -7,7 +7,7 @@ CALLER_DIR=$( pwd )
 cd "$( dirname "${BASH_SOURCE[0]}" )"
 
 if [[ "${VSCODE_ARCH}" == "x64" ]]; then
-  GITHUB_RESPONSE=$( curl --silent --location "https://api.github.com/repos/AppImage/pkg2appimage/releases/latest" )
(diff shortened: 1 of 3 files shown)


GitMensch:
@uli-heller: The changes in your fork [are in upstream now](https://github.com/AppImageCommunity/pkg2appimage/pull/545) [you may want to delete that fork], so I _guess_ this PR should be adjusted to:

* be rebased
* use the original URL
* uncomment the current disabled parts, just as you did

-> commit 1, may be the full PR here

* an extra commit to change the appimage check to drop the x86_64 part, to allow *later* to be built also for other environments
* as requested by @daiyam in the related issue, where you've noted that you've worked on this: have a separate workflow for the appimage

This may be part of either this PR or a separate one.

PR #1699, opened 2023-10-22
Add Render Glitch Workwaround to DOCS.MD

266-750Balloons:
Since VSCode/Codium 1.82.x, the Linux version, upon package updates, may open up to a mangled version of the interface missing text, making it completely unusable. This is an upstream Electron/Chromium issue with Mesa shaders (see microsoft/vscode#190437).
A work-around suggested in that issue is to delete the GPU cache with the following command:
```bash
rm -rf ~/.config/VSCodium/GPUCache
```
In case other users come across the same issue, I think it would be a good idea to address this in VSCodium documentation.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -17,6 +17,7 @@
 - [How do I press and hold a key and have it repeat in VSCodium?](#press-and-hold)
 - [How do I open VSCodium from the terminal?](#terminal-support)
   - [From Linux .tar.gz](#from-linux-targz)
+- [How do I fix rendering glitches such as text and/or the entire interface not appearing?](#rendering-glitches)
 - [How to build VSCodium](https://github.com/VSCodium/vscodium/blob/master/docs/build.md)
 
 ## <a id="disable-telemetry"></a>Getting all the Telemetry Out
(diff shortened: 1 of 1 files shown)


daiyam:
It's already in [Troubleshooting](https://github.com/VSCodium/vscodium/wiki/Troubleshooting)

daiyam:
Instead, we should add a link to the wiki page.

266-750Balloons:
I'll make that revision to the patch when I have the time.
In the meantime, I have to ask, why is there a separate DOCS.md from the the wiki?

daiyam:
@266-750Balloons I've reused your text in the wiki since it make things clearer. Thx

For the `DOCS.md`, not sure. Maybe, we should merge them.
And the `docs/build.md`, maybe, it should be renamed `CONTRIBUTING.md`.

daiyam:
@GitMensch @PalinuroSec @paulcarroty Should the `DOCS.md` be moved to the Wiki?

paulcarroty:
AFAIR Wiki access isn't flexible yet - rw access for all or only for maintainers, I don't like it.

paulcarroty:
> everything should be moved to the docs folder

and then it's possible to generate a static doc website using something like [mdBook](https://github.com/rust-lang/mdBook).

daiyam:
@266-750Balloons If you can update your PR and add your text to the file `/docs/troubleshooting.md`, we will merge it. Thx

PR #1715, opened 2023-11-10
feat(1.84.2): fix patch

daiyam:


Brli:
when doing `./prepare_vscode.sh under linux`
it returns
```
applying patch: ../patches/binary-name.patch
applying patch: ../patches/brand.patch
error: patch failed: src/vs/workbench/contrib/welcomeGettingStarted/common/gettingStartedContent.ts:184
error: src/vs/workbench/contrib/welcomeGettingStarted/common/gettingStartedContent.ts: patch does not apply
failed to apply patch ../patches/brand.patch
```

daiyam:
It's due to this issue https://github.com/microsoft/vscode/issues/197976

I'm thinking to the same as https://github.com/VSCodium/vscodium/blob/master/insider.json for for the stable branch.

Brli:
So, the suggestion would be to use a newer commit of vscode instead of the tagged one?

daiyam:
@Brli No, a `stable.json` would contains the correct commit id instead of looking at the tag to get the commit id (which currently points to an older version).

daiyam:
If by the end of the day, the tag isn't fixed then I will add the `stable.json`.

But it seems that something is screwed at GitHub since the tag is correct on some systems.

daiyam:
`stable.json` should be out soon #1728

daiyam:
The new version is out with https://github.com/VSCodium/vscodium/blob/master/stable.json

PR #1770, opened 2023-12-22
Add VSCodium WAPT package download in README.md

BazirClem:
Just referenced the VSCodium WAPT package in the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,13 @@ choco install vscodium
 choco install vscodium-insiders
 ```
 
+#### <a id="install-with-wapt"></a>Install with WAPT (Windows, Mac, Linux)
+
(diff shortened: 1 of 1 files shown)


daiyam:
@BazirClem Thanks for the PR, but here are my questions:
- Is WAPT open source?
- How the WAPT packages are created? Is it open source?
- Why the editor is `Microsoft`?

GitMensch:
It is proprietary and for administration in bigger networks (as far as I see especially for mixed-OS environments).
There was a free "community core" until WAPT 1.8, which is explicit said to be not used because of open CVEs, and the recent ones are completely proprietary.

I vote to not include WAPT in the README as this would mainly serve as "gratis advertising".

daiyam:
No answers?

PR #1795, opened 2024-01-30
fix: utilize APP_NAME in prepare_vscode.sh 

rizese:
Hi! While playing around with building, I noticed changing the `APP_NAME` in the `build/build.sh` script didn't actually change the name of the final executable and its bundles. So - here's a fix 👍 

This is tested ONLY in macOS

### DIFF
--- a/prepare_vscode.sh
+++ b/prepare_vscode.sh
@@ -58,6 +58,7 @@ set -x
 
 export ELECTRON_SKIP_BINARY_DOWNLOAD=1
 export PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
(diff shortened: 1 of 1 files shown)


daiyam:
Some of those `APP_NAME` are `vscodium` or `codium`.

For a fork, `updateUrl`, `downloadUrl` and lot of others place need to be customized.
Here, how I do it: https://github.com/zokugun/MrCode/blob/master/prepare.sh

The final naming of the executables are done in https://github.com/VSCodium/vscodium/blob/master/prepare_assets.sh
It's executed with `./build/build.sh -p`

rizese:
Thanks @daiyam !

PR #1796, opened 2024-01-30
disable linter for bash error SC2002

zpiatt:
This PR disables [https://www.shellcheck.net/wiki/SC2002](https://www.shellcheck.net/wiki/SC2002) in `build/build.sh` & `update_qualityjson.sh`.

`jq` should be able read files directly, however, if `cat` is the preference, disabling this error will minimize noise when linting/troubleshooting.


daiyam:
If you can make `jq` to output to the same file as the input, I'm all in that you update the specific commands.

PR #1817, opened 2024-02-27
[snap]: migrate to core22

soumyaDghosh:
Currently tested in

- [x]  Ubuntu 23.10
- [x] Kde Neon

Needed

- [ ] Ubuntu 20.04

Better if can be done

- [ ] Ubuntu 18.04
- [ ] Ubuntu 22.04

The error is gone. Hopefully it'll work fine

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ build/windows/msi/Files*.wxs
 build/windows/msi/Files*.wixobj
 stores/snapcraft/insider/*.snap
 stores/snapcraft/stable/*.snap
+stores/snapcraft/stable/codium_*.txt
--- a/stores/snapcraft/stable/snap/snapcraft.yaml
+++ b/stores/snapcraft/stable/snap/snapcraft.yaml
@@ -4,7 +4,7 @@ summary: Code editing. Redefined.
 description: |
   Binary releases of Code without branding/telemetry/licensing
 
-base: core20
+base: core22
 grade: stable
 confinement: classic
 compression: lzo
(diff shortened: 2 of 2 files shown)


soumyaDghosh:
@daiyam pinging for review and suggestions

soumyaDghosh:
@daiyam Any updates?

daiyam:
VSCode is still using `core20` so I don't know what to do...

daiyam:
Ok, they are still using `core20` but it seems that they will soon be moving to `core22`.

julian-alarcon:
Almost 1 year and still in use the old core20.

I cannot remove core20 because of this.

```sh
$ sudo snap remove core20
error: cannot remove "core20": snap "core20" is not removable: snap is being used by snaps codium
       and gnome-3-38-2004.
```

daiyam:
Visual Studio Code is still using `core20`

qwertychouskie:
VSCodium still using `core20` means I can't launch GUI apps from the built-in terminal, as it gives an undefined symbol error that seems related to glibc.  Is there a reason that VSCodium needs to use the same outdated/deprecated dependencies that MS's packaging uses?  For that matter, does `core20` even get security updates anymore, given that it's based on an EOL version of Ubuntu?

qwertychouskie:
The upstream snapcraft.yaml is here: https://github.com/microsoft/vscode/blob/main/resources/linux/snap/snapcraft.yaml

If someone familiar with the snap process adjusted this and made an upstream PR, perhaps it would get merged, and new upstream releases would actually use a modern core snap?

See also: https://github.com/microsoft/vscode/issues/241636

ReuelAlbert-Dev (commented):
need change

daiyam:
@soumyaDghosh Can you update the PR? If the snap works with any core version (even the 24), I will merge it.

I'm getting the error `Gtk: _gtk_style_provider_private_get_settings: assertion 'GTK_IS_STYLE_PROVIDER_PRIVATE (provider)' failed` on: 
- display about
- deleting a file

It would be a great help. Thx

qwertychouskie:
Upstream has this, though it seems stalled at the moment: https://github.com/microsoft/vscode/pull/298237

PR #1838, opened 2024-03-28
fix: Crashpad disabled regardless of `enable-crash-reporter` argv #1832

### LINKED ISSUE #1832: Crashpad disabled regardless of `enable-crash-reporter` argv
**Describe the bug**
Due to the patch introduced in #787 , regardless of the `enable-crash-reporter` argv, the crash-reporter is always disabled. 
- unless one start the program with `--crash-reporter-directory` present.
- the vscodium patch quits `configureCrashReporter` if there is no `product.appCenter`. There **always** isn't, is it?
- so there is **always** no electron crash minidumps for vscodium. Very frustrating for users who come to vscodium just for a pdb-available electron.

**Expected behavior**
1. The `enable-crash-reporter` argv is supposed to be the rightful controller over whether `configureCrashReporter` should be run. (check https://github.com/microsoft/vscode/blob/8a19756adf1b689e4f490a871d5067e6c54e3d01/src/main.js#L83C54-L83C75 ) 
2. If user `enable-crash-reporter`, the directory set for crashDumps (`%appdata/VSCodium/Crashpad` by default) should be initilized, containing an empty `settings.dat` and two directories `attachments` and `reports`. Minidumps are to be placed in the `reports` dir lest a crash happens someday.

**My Proposals**
1. The original problem that lead to #787 patch had been resolved by https://github.com/microsoft/vscode/commit/65475dc56e04da109cf3744987f2e00fafca914f . Should `submitURL` be empty or undefined, `uploadToServer` would be false, making it [acceptable](https://github.com/electron/electron/blob/main/docs/api/crash-reporter.md#crashreporterstartoptions) to `crashReporter.start()`. So we might as well remove the `crash-reporter.patch`.
2. Document the `--crash-reporter-directory <absolute-path>` cli option somewhere in the codium repository (based on infomation from [vscode](https://github.com/microsoft/vscode/wiki/Native-Crash-Issues)), so that users can have a better understanding of the crash telemetry, and quickly get to how to get dumps when electron crashes. 
+ From my opinion, this single cli option **enables** crash dump and **disables** crash telemetry at the same time, and is very worth recommending to our users. 
+ This infomation will also be **especially** helpful for users who have to stick on an earlier version of codium (permanently affected by the 787-patch!) due to compatibility reasons but also looking for minidumps.
3. For the current documents, the [mentioning](https://github.com/VSCodium/vscodium/blob/6253c391d182810740eed2622db79012e3c50e1d/docs/index.md?plain=1#L30C31-L30C62) of `telemetry.enableCrashReporter` has been obselete (and potentially misleading!) since the [one-time transition](https://code.visualstudio.com/updates/v1_49#_disabling-crash-reporter) in v1.49, and should be updated.

- [x] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/docs/index.md) and [the Troubleshooting page](https://github.com/VSCodium/vscodium/blob/master/docs/troubleshooting.md) and my issue is not mentioned there.

AgFlore:
fixes #1832


daiyam:
I've removed the crashpad sections in the docs since I don't think it's helpful for the general user.

PR #1866, opened 2024-05-06
Remove deprecated "unity-launch" flag from Linux desktop files

### LINKED ISSUE #1865: "--unity-launch" flag in Linux desktop file breaks file opening from the outside
**Describe the bug**
Since v1.89.0.24126, when Linux user tries to open a file from external program (e.g., file manager), VSCodium doesn't open a file but instead it simply launches a new VSCodium window.  
VSCodium's Linux desktop file still has a recently [removed](https://github.com/microsoft/vscode/commit/0da22a226000adc62cf7b8a83b045668fb8a9296) "[--unity-launch](https://github.com/microsoft/vscode/issues/202545)" flag.  
When I launch VSCodium with that flag, I receive such message in a terminal:

> Warning: 'unity-launch' is not in the list of known options, but still passed to Electron/Chromium.

And like any other non-existing flag, it breaks opening files from the outside (e.g., `codium --unity-launch launch ~/.bash_history`).

- [*] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

- [*] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/docs/index.md) and [the Troubleshooting page](https://github.com/VSCodium/vscodium/blob/master/docs/troubleshooting.md) and my issue is not mentioned there.

**To Reproduce**
1. On Linux desktop open any file manager (e.g., KDE Dolphin)
2. Right click on any text file
3. Choose "Open With VSCodium"
4. New VSCodium window without requested file is launched

**Expected behavior**
Requested file is opened in VSCodium

 - OS: Fedora 40
 - Architecture: x64
 - Version: v1.89.0.24126
 - App Manager: RPM
 - Sandboxed: no

**Additional context**
The problem is reproducible with different desktops (GNOME 46, KDE Plasma 6.0) so I presume it affects all Linux users. If possible, please make a new release with the fix to avoid duplicated reports for downstream packagers (Flathub, AUR etc.)

xalt7x:
Upstream [deprecated and removed](https://github.com/microsoft/vscode/commit/0da22a226000adc62cf7b8a83b045668fb8a9296) "unity-launch" flag.
This commit removes this redundant flag from Linux desktop files since it breaks file opening from the outside.

Closes #1865

### DIFF
--- a/src/insider/resources/linux/code.desktop
+++ b/src/insider/resources/linux/code.desktop
@@ -2,7 +2,7 @@
 Name=@@NAME_LONG@@
 Comment=Code Editing. Redefined.
 GenericName=Text Editor
-Exec=@@EXEC@@ --unity-launch %F
+Exec=@@EXEC@@ %F
 Icon=@@ICON@@
 Type=Application
 StartupNotify=false
--- a/src/stable/resources/linux/code.desktop
+++ b/src/stable/resources/linux/code.desktop
@@ -2,7 +2,7 @@
 Name=@@NAME_LONG@@
 Comment=Code Editing. Redefined.
 GenericName=Text Editor
-Exec=@@EXEC@@ --unity-launch %F
+Exec=@@EXEC@@ %F
 Icon=@@ICON@@
 Type=Application
 StartupNotify=false


daiyam:
@xalt7x Thanks for the catch!

wtechgo:
I'm on KDE Plasma 6 Manjaro fully updated and I still have this issue.

daiyam:
@wtechgo Yes, the AUR `vscodium-bin` doesn't have the fix since I can't push it (not co-maintainer on this one). I've emailed the maintainer but did get any answer (Just realized that I've sent him the wrong issue, sending another one).

wtechgo:
> @wtechgo Yes, the AUR `vscodium-bin` doesn't have the fix since I can't push it (not co-maintainer on this one). I've emailed the maintainer but did get any answer (Just realized that I've sent him the wrong issue, sending another one).

It's indeed `vscodium-bin` on my system. Thanks for sending the issue again.

wtechgo:
I just updated Manjaro, VScodium 1.90.0 came through and the issue is fixed.

PR #1919, opened 2024-06-07
fix: add vscode's smaller mipmaps to icons

EpicGaming11195:
ported vscode's icons' small mipmaps for medium size and smaller to vscodium's stable icons, so they won't look blurry when they're small

old:
image

new:
image

(i don't use github often, so i don't exactly know if fix is the right prefix for this commit)


daiyam:
I agree but I would prefer that the icons are generated by the script: https://github.com/VSCodium/vscodium/blob/master/icons/build_icons.sh

But you will need to modify the line
https://github.com/VSCodium/vscodium/blob/6a38a6e9db4afdb987f1bc688940965735432653/icons/build_icons.sh#L139

Then run:
- `./build/build.sh -o` to get VSCode source
- `./icons/build_icons.sh`

EpicGaming11195:
i use windows and barely know anything about linux or .sh files, so i wouldn't even know where to start with editing this script, but thanks for letting me know what exactly i will have to learn if/when i decide to do it

PR #1922, opened 2024-06-10
Add experimental riscv64 builds

kxxt:
With a performance issue(https://github.com/riscv-forks/electron/issues/1) fixed, I think the riscv64 port of electron is mostly ready for VSCodium. There are still some corner cases that lead to segfault but it's already suitable for usage on real riscv64 hardware.

This PR add riscv64 builds to vscodium.

Official electron doesn't support riscv64 so I used a fork maintained by myself.

BTW I noticed some of the patches modifies the compiled javascript files. I think it's cleaner to directly modify the typescript source file and compile ts files to javascrpt.

I haven't implemented riscv64 support in sysroot scripts so this PR only covers a basic riscv64 build, without DEB and RPM targets.

Here's a screenshot:

VSCodium 1.90 running on 64 core SG2042.

The window on the right shows using open-remote-ssh plugin to connect to x86_64 host from riscv64.

Screenshot_20_10_1

### DIFF
--- a/package_linux_bin.sh
+++ b/package_linux_bin.sh
@@ -22,6 +22,17 @@ if [[ "${VSCODE_ARCH}" == "ppc64le" ]]; then
   export VSCODE_SYSROOT_PREFIX='-glibc-2.28'
 fi
 
+if [[ "${VSCODE_ARCH}" == "riscv64" ]]; then
+  export VSCODE_ELECTRON_REPO='riscv-forks/electron-riscv-releases'
+  export ELECTRON_SKIP_BINARY_DOWNLOAD=1
+  export PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
+  ELECTRON_VERSION="v29.4.0"
+  # Look for releases here if electron version used by vscode changed
+  # https://github.com/riscv-forks/electron-riscv-releases/releases
+  export VSCODE_ELECTRON_TAG="${ELECTRON_VERSION}.riscv2"
+  echo "7244465fe0c1a6ac6e34fe765a9d90fe0017b1a6d3406fd6b8dd9f5d2c8c9df5 *electron-v29.4.0-linux-riscv64.zip" >> build/checksums/electron.txt 
+fi
+
 if [[ -d "../patches/${OS_NAME}/client/" ]]; then
   for file in "../patches/${OS_NAME}/client/"*.patch; do
     if [[ -f "${file}" ]]; then
--- a/patches/ppc64le-and-riscv64-support.patch
+++ b/patches/ppc64le-and-riscv64-support.patch
@@ -1,5 +1,5 @@
 diff --git a/build/azure-pipelines/linux/setup-env.sh b/build/azure-pipelines/linux/setup-env.sh
-index e42a6b12b1f..b31af12b086 100755
+index 9bfbf9ab41a..3064f783d8c 100755
 --- a/build/azure-pipelines/linux/setup-env.sh
 +++ b/build/azure-pipelines/linux/setup-env.sh
 @@ -7,6 +7,10 @@ if [ "$SYSROOT_ARCH" == "x64" ]; then
@@ -97,10 +97,33 @@ index 6f8144b0954..8a8195e9b66 100644
  
  BUILD_TARGETS.forEach(buildTarget => {
 diff --git a/build/gulpfile.vscode.js b/build/gulpfile.vscode.js
-index e1507e0424f..80ed8d91a41 100644
+index 3b1aeafd080..8363dae382c 100644
 --- a/build/gulpfile.vscode.js
 +++ b/build/gulpfile.vscode.js
-@@ -429,6 +429,8 @@ const BUILD_TARGETS = [
+@@ -339,11 +339,21 @@ function packageTask(platform, arch, sourceFolderName, destinationFolderName, op
+ 			all = es.merge(all, shortcut);
+ 		}
+ 
++		const electronOverride = {};
++		if (process.env.VSCODE_ELECTRON_REPO) {
++			// electron doesn't support riscv64 and ppc64le, override the repo with `VSCODE_ELECTRON_REPO`.
++			electronOverride.repo = process.env.VSCODE_ELECTRON_REPO;
++		}
++
++		if (process.env.VSCODE_ELECTRON_TAG) {
++			electronOverride.tag = process.env.VSCODE_ELECTRON_TAG;
++		}
++
+ 		let result = all
+ 			.pipe(util.skipDirectories())
+ 			.pipe(util.fixWin32DirectoryPermissions())
+ 			.pipe(filter(['**', '!**/.github/**'], { dot: true })) // https://github.com/microsoft/vscode/issues/116523
+-			.pipe(electron({ ...config, platform, arch: arch === 'armhf' ? 'arm' : arch, ffmpegChromium: false }))
++			.pipe(electron({ ...config, ...electronOverride, platform, arch: arch === 'armhf' ? 'arm' : arch, ffmpegChromium: false }))
+ 			.pipe(filter(['**', '!LICENSE', '!version'], { dot: true }));
+ 
+ 		if (platform === 'linux') {
+@@ -432,6 +442,8 @@ const BUILD_TARGETS = [
  	{ platform: 'linux', arch: 'x64' },
  	{ platform: 'linux', arch: 'armhf' },
  	{ platform: 'linux', arch: 'arm64' },
@@ -181,10 +204,10 @@ index 92f8065f262..b86b4dba2b8 100644
  	cmd.push(`-l${chromiumSysroot}/usr/lib`);
  	cmd.push(`-L${vscodeSysroot}/debian/libxkbfile1/DEBIAN/shlibs`);
 diff --git a/build/linux/debian/dep-lists.js b/build/linux/debian/dep-lists.js
-index bdb265b6fec..4695f335416 100644
+index d843c090063..3b869cbbe6d 100644
 --- a/build/linux/debian/dep-lists.js
 +++ b/build/linux/debian/dep-lists.js
-@@ -140,6 +140,85 @@ exports.referenceGeneratedDepsByArch = {
+@@ -139,6 +139,85 @@ exports.referenceGeneratedDepsByArch = {
          'libxkbfile1 (>= 1:1.1.0)',
          'libxrandr2',
          'xdg-utils (>= 1.0.2)'
@@ -272,10 +295,10 @@ index bdb265b6fec..4695f335416 100644
  //# sourceMappingURL=dep-lists.js.map
 \ No newline at end of file
 diff --git a/build/linux/debian/dep-lists.ts b/build/linux/debian/dep-lists.ts
-index 3d6c2eba6e9..b8987ea7579 100644
+index 4028370cd02..ff4df2f08cb 100644
 --- a/build/linux/debian/dep-lists.ts
 +++ b/build/linux/debian/dep-lists.ts
-@@ -140,5 +140,84 @@ export const referenceGeneratedDepsByArch = {
+@@ -139,5 +139,84 @@ export const referenceGeneratedDepsByArch = {
  		'libxkbfile1 (>= 1:1.1.0)',
  		'libxrandr2',
  		'xdg-utils (>= 1.0.2)'
@@ -401,7 +424,7 @@ index feca7d3fa9d..f075ebcba75 100644
      console.log(`Fetching ${expectedName} for ${triple}`);
      const checksumSha256 = getVSCodeSysrootChecksum(expectedName);
 diff --git a/build/linux/debian/install-sysroot.ts b/build/linux/debian/install-sysroot.ts
-index aa417dcc722..02b1cc4e057 100644
+index aa417dcc722..58f0eeefeff 100644
 --- a/build/linux/debian/install-sysroot.ts
 +++ b/build/linux/debian/install-sysroot.ts
 @@ -82,7 +82,10 @@ async function fetchUrl(options: IFetchOptions, retries = 10, retryDelay = 1000)
@@ -409,10 +432,10 @@ index aa417dcc722..02b1cc4e057 100644
  		const version = '20240129-253798';
  		try {
 -			const response = await fetch(`https://api.github.com/repos/Microsoft/vscode-linux-build-agent/releases/tags/v${version}`, {
-+				// if VSCODE_SYSROOT_REPO and VSCODE_SYSROOT_VERSION are set, use those instead
-+				const repo = process.env['VSCODE_SYSROOT_REPO'] ?? 'Microsoft/vscode-linux-build-agent';
-+				const actualVersion = process.env['VSCODE_SYSROOT_VERSION'] ?? version;
-+				const response = await fetch(`https://api.github.com/repos/${repo}/releases/tags/v${actualVersion}`, {
++			// if VSCODE_SYSROOT_REPO and VSCODE_SYSROOT_VERSION are set, use those instead
++			const repo = process.env['VSCODE_SYSROOT_REPO'] ?? 'Microsoft/vscode-linux-build-agent';
++			const actualVersion = process.env['VSCODE_SYSROOT_VERSION'] ?? version;
++			const response = await fetch(`https://api.github.com/repos/${repo}/releases/tags/v${actualVersion}`, {
  				headers: ghApiHeaders,
  				signal: controller.signal as any /* Typings issue with lib.dom.d.ts */
  			});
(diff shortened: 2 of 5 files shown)


lex-ibm:
Ups... I think we were working on more or less the same...

lex-ibm:
Should we coordinate and have this be a single PR for both RISCV64 and PPC64LE?

daiyam:
> Should we coordinate and have this be a single PR for both RISCV64 and PPC64LE?

We should split the `ppc64le-and-riscv64-support.patch`.

build.sh
```diff
@@ -14,6 +14,11 @@ if [[ "${SHOULD_BUILD}" == "yes" ]]; then
 
   export NODE_OPTIONS="--max-old-space-size=8192"
 
+  # Some patches(e.g. remove-mangle.patch) causes unused variables and parameters
+  sed -ri '/noUnused(Local|Parameter)s/d' build/tsconfig.json
+  # Recompile ts files in ./build since the patches modifies them
+  yarn --cwd build compile
```
daiyam, line 20: No needed since the patches also modify the `.js`

patches/remove-mangle.patch
```diff
@@ -1,37 +1,10 @@
-diff --git a/build/lib/compilation.js b/build/lib/compilation.js
```
daiyam, line 1: Do not modify that patch.

patches/ppc64le-and-riscv64-support.patch
```diff
@@ -372,10 +393,10 @@ index 4bec13d6e86..3debda1224a 100644
  			Platform::DarwinARM64 => "DarwinARM64",
  			Platform::WindowsX64 => "WindowsX64",
 diff --git a/cli/src/util/prereqs.rs b/cli/src/util/prereqs.rs
-index 20a5bc94b37..50678354245 100644
+index 5d4953acbb2..f22ff60e391 100644
```
daiyam, line 396: Not needed

patches/ppc64le-and-riscv64-support.patch
```diff
@@ -409,10 +430,10 @@ index 079557869e3..c057746b870 100644
  
  if [ "$OS_ID" != "alpine" ]; then
 diff --git a/src/vs/platform/extensionManagement/common/extensionManagement.ts b/src/vs/platform/extensionManagement/common/extensionManagement.ts
-index 9dae82eba07..94246285b26 100644
+index 183f2582871..2e6614cad59 100644
```
daiyam, line 433: Not needed

lex-ibm:
> > Should we coordinate and have this be a single PR for both RISCV64 and PPC64LE?
> 
> We should split the `ppc64le-and-riscv64-support.patch`.

Roger that!

patches/linux/use-node16.patch
```diff
@@ -1,19 +1,27 @@
...
- // Based on https://source.chromium.org/chromium/chromium/src/+/main:build/linux/sysroot_scripts/install-sysroot.py.
+diff --git a/build/lib/fetch.ts b/build/lib/fetch.ts
+index dc1de777e04..e62db40bd75 100644
+--- a/build/lib/fetch.ts
++++ b/build/lib/fetch.ts
+@@ -9,6 +9,7 @@ import * as log from 'fancy-log';
+ import * as ansiColors from 'ansi-colors';
+ import * as crypto from 'crypto';
+ import * as through2 from 'through2';
++const fetch: any = require('node-fetch');
```
daiyam, line 9: Please revert the changes on the file

patches/fix-build-linux.patch
```diff
@@ -18,20 +18,6 @@ index 8c2b62f..843e8d4 100644
 -	const buildRpmTask = task.define(`vscode-linux-${arch}-build-rpm`, buildRpmPackage(arch));
 +	const buildRpmTask = task.define(`vscode-linux-${arch}-build-rpm`, task.series(prepareRpmTask, buildRpmPackage(arch)));
  	gulp.task(buildRpmTask);
-diff --git a/build/linux/dependencies-generator.js b/build/linux/dependencies-generator.js
```
daiyam, line 21: Please revert the changes on the file

patches/ext-from-gh.patch
```diff
@@ -1,22 +1,15 @@
-diff --git a/build/lib/builtInExtensions.js b/build/lib/builtInExtensions.js
-index 1b0adc4..b595123 100644
---- a/build/lib/builtInExtensions.js
-+++ b/build/lib/builtInExtensions.js
```
daiyam, line 4: Please revert the changes on the file

kxxt:
> > Should we coordinate and have this be a single PR for both RISCV64 and PPC64LE?
> 
> We should split the `ppc64le-and-riscv64-support.patch`.

I think it might be cleaner to keep one patch. If we split the patch, the ppc64 and riscv64 patches both modifies the same line in a file, which will lead to a conflict. WDYT?

kxxt:
> Should we coordinate and have this be a single PR for both RISCV64 and PPC64LE?

Does the PPC64LE has an electron fork that publishes electron binaries? If so I could add it in this PR?

build.sh
kxxt: Done

lex-ibm:
> > Should we coordinate and have this be a single PR for both RISCV64 and PPC64LE?
> 
> Does the PPC64LE has an electron fork that publishes electron binaries? If so I could add it in this PR?

We don't publish the binaries, but we use the `ELECTRON_MIRROR` variable to set the location for it. The PR that I opened only makes the changes needed to enable the build to happen. Is up to the user to specify where that electron binary comes from.

kxxt:
> > > Should we coordinate and have this be a single PR for both RISCV64 and PPC64LE?
> > 
> > 
> > Does the PPC64LE has an electron fork that publishes electron binaries? If so I could add it in this PR?
> 
> We don't publish the binaries, but we use the `ELECTRON_MIRROR` variable to set the location for it. The PR that I opened only makes the changes needed to enable the build to happen. Is up to the user to specify where that electron binary comes from.

Oh sorry, I didn't see your PR. I thought you haven't opened it. About the patches, do you think it's more convenient to split riscv64 and ppc64 or keep one patch?  I think splitting the patch doesn't actually solve the conflict but makes it more complex because we need to stack a patch on top of another.

lex-ibm:
> I think it might be cleaner to keep one patch. If we split the patch, the ppc64 and riscv64 patches both modify the same line in a file, which will lead to a conflict. WDYT?

I kinda agree, I tried splitting the patch and the only way I got it to work was if I based one on top of the other. Also, in the other PR there are some changes to how to get the vscode-sysroot that make it so that we don't need to check for a specific arch.

While developing it I was trying to make it so that porting to RISCV64 was more or less easy, but never tried to build it.

lex-ibm:
>  think splitting the patch doesn't actually solve the conflict but makes it more complex because we need to stack a patch on top of another.

Agree, took me a couple of hours just to split the changes and always diffing everything to make sure I didn't mess anything up. I kinda feel like this could be a single PR or at least we should both coordinate on how to get them both in.

I do recommend you take the patch to `install-sysroot.js`, that makes it easier if the tag changes or we want to use another repo.

kxxt:
> I do recommend you take the patch to `install-sysroot.js`, that makes it easier if the tag changes or we want to use another repo.

I agree with you. There's no need to maintain two methods of using a custom electron repo.

kxxt:
Another way is to decommonize the intermediate vscode artifact, that is, building an intermediate vscode artifact for each architecture. This way we can apply different patches to different architectures. But I don't think it worth the effort for just two architectures that need patching. This method might make more sense if we have patches for four or more architectures that need patching.

lex-ibm:
> This method might make more sense if we have patches for four or more architectures that need patching.

As long as no one asks me to build for `s390x` I'm fine with that.

Let's get started then! I'm in `GMT-5`, so this is probably the best time to contact me.

I see that you replace the `electron.txt` checksums. That would collide with my needs, since that checksums file wouldn't have the sha256 for our binary.

package_linux_bin.sh
```diff
@@ -16,6 +16,21 @@ cd vscode || { echo "'vscode' dir not found"; exit 1; }
...
+  export ELECTRON_SKIP_BINARY_DOWNLOAD=1
+  export PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
+  # Official electron doesn't support riscv64, adding the checksums of forked electron here
+  # Electron riscv fork might do multiple builds for a same version,
+  # the tag is in the format of v<version>.riscv<build_number>.
+  # In order to be compatible with upstream release scheme, riscv fork also re-upload the latest
+  # RISC-V release to the upstream release tag(v<version>). 
+  # This might be problematic if there is an old electron zip cached locally.
+  rm -f ~/.cache/electron/*/electron-*-linux-riscv64.zip
+  ELECTRON_VERSION="v$(yarn config get target)"
```
daiyam, line 29: It would be better to have specific version number and its checksum.
This way, we know for sure which electron version has been used for a specific VSCodium version.
It allows us to be able to reproduce that version if needed.

patches/ppc64le-and-riscv64-support.patch
```diff
@@ -97,10 +97,29 @@ index 6f8144b0954..8a8195e9b66 100644
...
 +++ b/build/gulpfile.vscode.js
-@@ -429,6 +429,8 @@ const BUILD_TARGETS = [
+@@ -339,11 +339,17 @@ function packageTask(platform, arch, sourceFolderName, destinationFolderName, op
+ 			all = es.merge(all, shortcut);
+ 		}
+ 
++		const electronOverride = {};
++		if (platform === 'linux' && arch === 'riscv64') {
++			// electron doesn't support riscv64, so a fork is used here.
++			electronOverride.repo = 'riscv-forks/electron-riscv-releases';
```
lex-ibm, line 110: This can be achieved with `ELECTRON_MIRROR`, can it not?

kxxt: Yes.

package_linux_bin.sh
kxxt: > It would be better to have specific version number and its checksum. This way, we know for sure which electron version has been used for a specific VSCodium version. It allows us to be able to reproduce that version if needed.

Yes, I agree with you. I could use a specific version like v29.4.2.riscv2 but it would require me to make more PRs in the future to maintain the riscv64 revision so I didn't take this approach on the first try.

kxxt:
> > This method might make more sense if we have patches for four or more architectures that need patching.
> 
> As long as no one asks me to build for `s390x` I'm fine with that.
> 
> Let's get started then! I'm in `GMT-5`, so this is probably the best time to contact me.
> 
> I see that you replace the `electron.txt` checksums. That would collide with my needs, since that checksums file wouldn't have the sha256 for our binary.

It wouldn't collide with your needs. The riscv64 and ppc64le pipelines are ran separately so replacing it in riscv64 shouldn't affect it's usage in ppc64le.

patches/ppc64le-and-riscv64-support.patch
daiyam: Not sure. `ELECTRON_MIRROR` seems to be a `electron-builder` variable.
VSCode seems to use `electronRepository` in the `product.json`

https://github.com/microsoft/vscode/blob/0fd9a006f7abcb48ef1e5a9f9ceeca33e1d1daaa/build/lib/electron.ts#L201

kxxt: > This can be achieved with `ELECTRON_MIRROR`, can it not?

@lex-ibm I didn't see `ELECTRON_MIRROR` anywhere in your PR? Where did you implemented it in your PR?

kxxt: > This can be achieved with `ELECTRON_MIRROR`, can it not?

@lex-ibm I didn't see `ELECTRON_MIRROR` anywhere in your PR? Where did you implemented it in your PR?

kxxt: Oh the network glitch caused me to send the comment twice. Sorry.

daiyam:
> It wouldn't collide with your needs. The riscv64 and ppc64le pipelines are ran separately so replacing it in riscv64 shouldn't affect it's usage in ppc64le.

It will be better if you could just patch the `electron.txt`

daiyam:
> > It wouldn't collide with your needs. The riscv64 and ppc64le pipelines are ran separately so replacing it in riscv64 shouldn't affect it's usage in ppc64le.
> 
> Is it ok with you if the binaries for the electron binary and it's sha256 change without VSCodium knowing? I kinda feel like that might be a bit risky.

I agree

patches/ppc64le-and-riscv64-support.patch
kxxt: Resolved by using a `VSCODE_ELECTRON_REPO` environment variable for the electron repo.

daiyam:
> I actually don't replace the checksums but append to it. @daiyam Does it look good to you? I consider it as a method of "patch".

No, I prefer to have version and checksum in clear text in the repo. There are several benefit to do this:
- every body can look at the code and know which version has been used
- reproducible binaries
- the changes of version are tracked

We do the same for VSCode ([stable.json](https://github.com/VSCodium/vscodium/blob/master/stable.json))

lex-ibm:
I don't know if it helps, but I do remember seeing an environment variable to change the electron version, I just don't remember where. I'll look for it.
With that I think you could use `ELECTRON_MIRROR` and whatever the other variable is and all that would be needed is the patch for `electron.txt`.

kxxt:
> > I actually don't replace the checksums but append to it. @daiyam Does it look good to you? I consider it as a method of "patch".
> 
> No, I prefer to have version and checksum in clear text in the repo. There are several benefit to do this:
> 
>     * every body can look at the code and know which version has been used
> 
>     * reproducible binaries
> 
>     * the changes of version are tracked
> 
> 
> We do the same for VSCode ([stable.json](https://github.com/VSCodium/vscodium/blob/master/stable.json))

Thanks for clarifying. I have squashed the commits and rebased my PR. Will mark ready for review once the CI passes here: https://github.com/kxxt/vscodium/actions/runs/9468203671

lex-ibm:
[Found it!](https://github.com/electron/get/blob/main/src/artifact-utils.ts) Can even specify a custom file name! Does that help?

daiyam:
> @daiyam I write the clear text electron checksum in `package_linux_bin.sh`. Do you think I need to create a patch file for it or it's fine to maintain the checksum in `package_linux_bin.sh`?

You don't need a patch

kxxt:
> > @daiyam I write the clear text electron checksum in `package_linux_bin.sh`. Do you think I need to create a patch file for it or it's fine to maintain the checksum in `package_linux_bin.sh`?
> 
> You don't need a patch

Then I think the PR is ready for review now. The CI is green here(for riscv64 and ppc64le): https://github.com/kxxt/vscodium/actions/runs/9468203671/job/26084811222

daiyam:
> [Found it!](https://github.com/electron/get/blob/main/src/artifact-utils.ts) Can even specify a custom file name! Does that help?

@lex-ibm VSCode seems to use the package [@vscode/gulp-electron](https://www.npmjs.com/package/@vscode/gulp-electron) to download electron (https://github.com/microsoft/vscode/blob/5adc4dcb025aac567c55100c4f2777614644de89/build/lib/electron.ts#L208)

package_linux_bin.sh
```diff
@@ -22,6 +22,17 @@ if [[ "${VSCODE_ARCH}" == "ppc64le" ]]; then
   export VSCODE_SYSROOT_PREFIX='-glibc-2.28'
 fi
 
+if [[ "${VSCODE_ARCH}" == "riscv64" ]]; then
+  export VSCODE_ELECTRON_REPO='riscv-forks/electron-riscv-releases'
+  export ELECTRON_SKIP_BINARY_DOWNLOAD=1
+  export PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
+  ELECTRON_VERSION="v$(yarn config get target)"
```
daiyam, line 29: `ELECTRON_VERSION="v29.4.0"` should be enough

kxxt: Fixed

patches/ppc64le-and-riscv64-support.patch
```diff
@@ -97,10 +97,33 @@ index 6f8144b0954..8a8195e9b66 100644
...
 +++ b/build/gulpfile.vscode.js
-@@ -429,6 +429,8 @@ const BUILD_TARGETS = [
+@@ -339,11 +339,21 @@ function packageTask(platform, arch, sourceFolderName, destinationFolderName, op
+ 			all = es.merge(all, shortcut);
+ 		}
+ 
++		const electronOverride = {};
++		if (process.env.VSCODE_ELECTRON_REPO) {
++			// electron doesn't support riscv64 and ppc64le, override the repo with `VSCODE_ELECTRON_REPO`.
++			electronOverride.repo = process.env.VSCODE_ELECTRON_REPO;
```
daiyam, line 110: Why not directly modifying the `config` variable?

daiyam:
LGTM

@kxxt Is it ok for you?

daiyam:
@lex-ibm Do you have any feedback?

kxxt:
> LGTM
> 
> @kxxt Is it ok for you?

Yes

lex-ibm:
LGTM! Awesome job!

daiyam:
@kxxt Thanks!

kxxt:
Thanks both of you for helping me to get this PR merged!

About automatically updating the riscv64 electron version and checksum. I think we can implement it in a script and run it in the *-spearhead workflows.

PR #1926, opened 2024-06-11
Add Github Enterprise support

lex-ibm:
Adds support for using this repo in a private Github Enterprise instance.

### DIFF
--- a/check_tags.sh
+++ b/check_tags.sh
@@ -3,17 +3,22 @@
 
 set -e
 
-if [[ -z "${GITHUB_TOKEN}" ]]; then
+if [[ -z "${GH_TOKEN}" ]] && [[ -z "${GITHUB_TOKEN}" ]] && [[ -z "${GH_ENTERPRISE_TOKEN}" ]] && [[ -z "${GITHUB_ENTERPRISE_TOKEN}" ]]; then
   echo "Will not build because no GITHUB_TOKEN defined"
   exit 0
+else
+  GITHUB_TOKEN="${GH_TOKEN:-${GITHUB_TOKEN:-${GH_ENTERPRISE_TOKEN:-${GITHUB_ENTERPRISE_TOKEN}}}}"
 fi
+
+# Support for GitHub Enterprise
+GH_HOST="${GH_HOST:-github.com}"
 
 APP_NAME_LC="$( echo "${APP_NAME}" | awk '{print tolower($0)}' )"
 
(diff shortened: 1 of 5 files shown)


lex-ibm:
I did not include the `SHOULD_BUILD_TAR="${SHOULD_BUILD_TAR:-no}"` changes. I don't think those affect the current build process but given the script is not meant to be used like that I don't see the case.

check_tags.sh
```diff
@@ -3,17 +3,20 @@
 
 set -e
 
-if [[ -z "${GITHUB_TOKEN}" ]]; then
+if [[ -z "${GH_TOKEN}" ]] && [[ -z "${GITHUB_TOKEN}" ]] && [[ -z "${GH_ENTERPRISE_TOKEN}" ]] && [[ -z "${GITHUB_ENTERPRISE_TOKEN}" ]]; then
```
daiyam, line 6: Is it missing the `else` statement like in `update_version.sh`?

daiyam:
> I did not include the `SHOULD_BUILD_TAR="${SHOULD_BUILD_TAR:-no}"` changes. I don't think those affect the current build process but given the script is not meant to be used like that I don't see the case.

Yep, it doesn't affect the VSCodium's build process. So it's up to you.

check_tags.sh
lex-ibm: It was. Changed it to match.

lex-ibm:
I'll change my pipeline instead. I don't want to change how `check_tags.sh` works other than adding GitHub Enterprise support.

daiyam:
LGTM

daiyam:
OH! A last thought, wouldn't `GITHUB_HOST` be better than `GH_HOST`?

daiyam:
> OH! A last thought, wouldn't `GITHUB_HOST` be better than `GH_HOST`?

Can you make the change?

lex-ibm:
> > OH! A last thought, wouldn't `GITHUB_HOST` be better than `GH_HOST`?
> 
> Can you make the change?

`GH_HOST` comes from the [GitHub CLI documentation](https://cli.github.com/manual/gh_help_environment).

lex-ibm:
Changing `GH_HOST` to `GITHUB_HOST` would mean changing the calls to `gh` adding the `--hostname` flag.

daiyam:
@lex-ibm Thanks

lex-ibm:
@daiyam thank you! I'll work on getting those electron binaries out.

Issue #1947, opened 2024-07-06
Support for Node.js v20 on Linux


Suyun114:
Currently, VSCodium builds successfully with Node.js v20 on Windows and macOS, but it crashes when building on Linux. We are limited to using Node.js v18 on Linux, which is inconvenient.

The official VSCode project supports Node.js v20 for Linux builds. Can we get the same support in VSCodium?

Suyun114:
I tried to fix it on #1949.

daiyam:
We provide a wider support of systems since we haven't migrated to Node-v20.

You can use tool like [nvm](https://github.com/nvm-sh/nvm) to easily switch to Node-v18

daiyam:
I will make some changes so that the patches for node16 aren't not applied by default.

PR #1949, opened 2024-07-07
fix(linux): fix build with node-v20

Suyun114:


Suyun114:
Fix #1947.

daiyam:
The PR will stop us to support older linux.

PR #1950, opened 2024-07-07
feat: update to 1.91.0

daiyam:


apainintheneck:
FYI: v1.91.0 doesn't seem to work on older macs running Catalina or Big Sur. See https://github.com/microsoft/vscode/issues/219893

daiyam:
@apainintheneck Thanks for the update. It seems that it's due to the version of macos of their build agent. I won't change ours so it should be fine (I had the same issue before, #1847).
But their fix is to use parallels??? WTF!

apainintheneck:
Ah, good to know. I didn't realize that you'd run into that issue before and that it had been resolved. It's understandable that a buggy build for an ancient OS version doesn't get noticed but at the same time it might be a sign of more problems with their build pipeline configuration.

daiyam:
@apainintheneck You can find the temp installers at https://github.com/VSCodium/vscodium/actions/runs/9836298705 in the artifacts (at the bottom of the page) 

It's working on Big Sur x86 but the issue was with Big Sur on arm...

PR #1959, opened 2024-07-16
fix(electron): fix electron version of riscv64

Suyun114:


package_linux_bin.sh
```diff
@@ -26,11 +26,12 @@ if [[ "${VSCODE_ARCH}" == "riscv64" ]]; then
   export VSCODE_ELECTRON_REPO='riscv-forks/electron-riscv-releases'
   export ELECTRON_SKIP_BINARY_DOWNLOAD=1
   export PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
-  ELECTRON_VERSION="v29.4.0"
+  ELECTRON_VERSION="v30.2.0"
```
kxxt, line 29: The version here should match `.yarnrc` to avoid native node module ABI breakage.

Upstream is still using v29.4.0. Is there any specific reason to bump to v30.2.0?
https://github.com/microsoft/vscode/blob/1.91.1/.yarnrc#L2

kxxt (changes requested):

PR #1970, opened 2024-07-22
feat: prepare for 1.92

daiyam:


kxxt:
Hi @daiyam, In case you haven't noticed it, I opened PR  https://github.com/daiyam/vscodium/pull/2 targeting daiyam:feat-192 to fix the riscv CI failure.

PR #2041, opened 2024-09-24
fix: check both `arm64` and `aarch64`

Suyun114:


daiyam:
@Suyun114 Can you confirm it's to build on Linux+ARM?

PR #2055, opened 2024-10-10
feat(1.95): update patches

daiyam:


daiyam:
@lex-ibm Hi, I have disabled ppc64 build.
Because the dependency `@vscode/policy-watcher` is using the gcc's flag `-std=gnu++20`, so I have updated to `gcc-10`.
I'm not sure how to update the `sysroot`, if you could do it, it would be great.

lex-ibm:
Roger that, will get this out asap

lex-ibm:
@daiyam I saw the `focal-ppc64le` build failed, but I was able to build it in my local. Was qemu enabled for that job?

lex-ibm:
I believe [this line](https://github.com/VSCodium/vscode-linux-build-agent/blob/bd48b4709943e33404d9db89dbb18511ef581f19/.github/workflows/build.yml#L31) is missing `focal-ppc64le`

daiyam:
> I believe [this line](https://github.com/VSCodium/vscode-linux-build-agent/blob/bd48b4709943e33404d9db89dbb18511ef581f19/.github/workflows/build.yml#L31) is missing `focal-ppc64le`

yep!

daiyam:
From https://github.com/VSCodium/vscodium/actions/runs/11328290273/job/31501917695, it's still using the gcc of the sysroot (gcc-8.5), define in the patch (`export CC=$VSCODE_SYSROOT_DIR/powerpc64le-linux-gnu/bin/powerpc64le-linux-gnu-gcc`).
In #2059, I'm trying to see if we can avoid the sysroot since riscv64 do compile without it.

PR #2059, opened 2024-10-14
fix: build ppc64

daiyam:


lex-ibm:
I have some work that I can't put to the side, how long can this wait until it becomes a MUST? I'm pretty sure I'm going to be flooded with messages if there is an update and there isn't a PPC64LE REH binary hahaha.

daiyam:
> I have some work that I can't put to the side, how long can this wait until it becomes a MUST? 

Well, I can disable a REH build with a GA variable like `DISABLE_INSIDER_LINUX_REH_PPC64=yes`.
So it's not a blocker for me. Only that the PPC64 release will be delayed.

> I'm pretty sure I'm going to be flooded with messages if there is an update and there isn't a PPC64LE REH binary hahaha.

Yep!


This PR tries to build RISCV but to no avail...
Maybe building a new sysroot will be faster...


> ```
> npm error cc: error: unrecognized command-line option ‘-mminimal-toc’
> ```
> 
> That's weird... why wouldn't it be using gcc for powerpc64le?

`cc` it's part of gcc but `-mminimal-toc` is only available since gcc-10. I don't know why it don't use it.


Another solution would be to swap the problematic `-std=gnu++20` with `-std=gnu++2a` in `@vscode/policy-watcher` during the `npm install`. But I don't know if that possible.
MS seems to use private repo where that lib has been pre-built with [azure pipeline](https://github.com/microsoft/vscode-policy-watcher/blob/main/azure-pipelines.yml).

daiyam:
closing in favor of #2077

lex-ibm:
@daiyam I wanted to release this under the [ppc64le/build-scripts](https://github.com/ppc64le/build-scripts) repo, but anyway.

I've created a repo with the build scripts that I use for electron.
- [lex-ibm/electron-ppc64le-build-scripts](https://github.com/lex-ibm/electron-ppc64le-build-scripts)

I want to add a GH Actions pipeline but for the moment I'm having some issues cross-compiling from x64 and arm64 to ppc64le.

I'll probably release some binaries for 32.2.5 that I personally built. I'll sign those binaries with my gpg key. I know that doesn't remove the risk of a supply chain attack, but is the best I can do at the moment until I figure out the cross-compile issues.

daiyam:
> @daiyam I wanted to release this under the [ppc64le/build-scripts](https://github.com/ppc64le/build-scripts) repo

No problem.

> I want to add a GH Actions pipeline but for the moment I'm having some issues cross-compiling from x64 and arm64 to ppc64le.

Have you look at https://github.com/riscv-forks/electron-riscv-releases or https://github.com/darkyzhou/electron-loong64?

lex-ibm:
I have, but the current issue I'm having is with `v8`. I'll probably create the GH Action and let it run until it fails. I've already uploaded my current patch to the `next` branch.
Side note. If there is any issue with PPC64LE VSCodium that requires my immediate attention, my email is in the headers of that build script, you can use that to contact me.
I also want to touch on what the issue with the GCC 8.5 was. I'm using a different setup (a custom `rhel8-ppc64le` instead of `focal-ppc64le`, should I upload that in case anyone wants it or is it too much clutter?) and I'm not sure what the fix was.

PR #2089, opened 2024-11-04
Upgrade `macos-12` to `macos-13`

haohanyang:
The x64 runner image `macos-12` is deprecated according to https://github.com/actions/runner-images?tab=readme-ov-file#available-images 

The PR upgrades the it to a newer `macos-13`

The latest x64 image is `macos-latest-large` or `macos-14-large` which seems to be a paid feature

### DIFF
--- a/.github/workflows/stable-macos.yml
+++ b/.github/workflows/stable-macos.yml
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - runner: macos-12
+          - runner: macos-13
             vscode_arch: x64
           # - runner: macos-14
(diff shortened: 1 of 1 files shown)


daiyam:
Sorry, I will wait until the latest moment since building on newer macos might result on app that might not work on older macos. https://github.com/VSCodium/vscodium/issues/1847

daiyam:
By the way, GitHub have emailed me on the 16th about this change. So that's why I had already an opinion about it.

pheiduck:
> The latest x64 image is `macos-latest-large` or `macos-14-large` which seems to be a paid feature

There is a free image `macos-15-intel`
See #2605

PR #2095, opened 2024-11-06
fix appimage build by setting ARCH env

### LINKED ISSUE #2003: No appimage builds since 1.92
**Describe the bug**
No appimage builds provided in github releases.

- [*] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

- [*] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/docs/index.md) and [the Troubleshooting page](https://github.com/VSCodium/vscodium/blob/master/docs/troubleshooting.md) and my issue is not mentioned there.

**To Reproduce**
1. Attempt to download appimage from releases.
2. See there is no appimage build.

**Expected behavior**
There is an appimage build available in github releases.

**Screenshots**
N/A

N/A

**Additional context**
appimages may not be important to most, but in some situations it is the only way to get an app working on a platform.

KnowZero:
This patch restores the appimage builds by setting ARCH=x86_64 env to fix the building error.

Fixes #2003

### DIFF
--- a/build/linux/appimage/build.sh
+++ b/build/linux/appimage/build.sh
@@ -19,6 +19,8 @@ if [[ "${VSCODE_ARCH}" == "x64" ]]; then
 
   chmod +x ./pkg2appimage.AppImage
 
+  export ARCH=x86_64
+
   ./pkg2appimage.AppImage --appimage-extract && mv ./squashfs-root ./pkg2appimage.AppDir
 
   # add update's url


daiyam:
Thanks, I will merge the PR after few chnages:
- can you it before line 40
- can you add a comment of why it's here

KnowZero:
@daiyam done, does it look fine now?

daiyam:
The `AppImage` is available.

PR #2099, opened 2024-11-09
add s390x support for REH

andreasgrapentin:
This PR extends VSCodium by adding s390x (IBM Mainframe / System Z) to the list of valid targets for the remote extension host. 

This is mostly straightforward and very similar to the other supported architectures, with one difference that is required in the dependency tree of vscode, where web-tree-sitter needs to be updated to work around an endianness issue.

The CI components as used in this PR are currently still built in my forked version of vscode-linux-build-agent, but I have created a separate PR for that repository here: https://github.com/VSCodium/vscode-linux-build-agent/pull/24, if that is merged, then the paths used here can be updated accordingly.

I have been using a manually built and installed version of the REH for a few weeks now, and have not noticed any issues or regressions.

### DIFF
--- a/.github/workflows/insider-linux.yml
+++ b/.github/workflows/insider-linux.yml
@@ -279,6 +279,8 @@ jobs:
         - slug: LOONG64
           vscode_arch: loong64
           npm_arch: loong64
+        - vscode_arch: s390x
+          npm_arch: s390x
     env:
       BUILD_SOURCEVERSION: ${{ needs.compile.outputs.BUILD_SOURCEVERSION }}
       DISABLED: ${{ vars[format('DISABLE_INSIDER_LINUX_REH_{0}', matrix.slug)] }}
--- a/.github/workflows/insider-linux.yml
+++ b/.github/workflows/insider-linux.yml
@@ -279,7 +279,8 @@ jobs:
         - slug: LOONG64
           vscode_arch: loong64
           npm_arch: loong64
-        - vscode_arch: s390x
+        - slug: S390X
+          vscode_arch: s390x
           npm_arch: s390x
     env:
       BUILD_SOURCEVERSION: ${{ needs.compile.outputs.BUILD_SOURCEVERSION }}
--- a/patches/linux/arch-4-s390x.patch
+++ b/patches/linux/arch-4-s390x.patch
@@ -23,7 +23,7 @@ index 0b5f38c..9f3b02b 100644
 +++ b/build/checksums/vscode-sysroot.txt
 @@ -7 +7,2 @@
  fa8176d27be18bb0eeb7f55b0fa22255050b430ef68c29136599f02976eb0b1b  powerpc64le-linux-gnu-glibc-2.28.tar.gz
-+c16511a2466a78d22c2b33d89f825fcb3fcd085ebfd099bb0c852ed730e0e78a  s390x-linux-gnu-glibc-2.28.tar.gz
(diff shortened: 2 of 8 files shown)


.github/workflows/insider-linux.yml
```diff
@@ -279,6 +279,8 @@ jobs:
         - slug: LOONG64
           vscode_arch: loong64
           npm_arch: loong64
```
andreasgrapentin, line 281: I missed the introduction of the `slug` field when rebasing my changes to master earlier today. I will add a slug with the next update of the PR.

patches/linux/arch-4-s390x.patch
```diff
@@ -0,0 +1,465 @@
...
+       "version": "3.0.1",
+diff --git a/package.json b/package.json
+index 7e0de18..4ceaded 100644
+--- a/package.json
++++ b/package.json
+@@ -226,7 +226,10 @@
+     },
+     "@parcel/watcher@2.1.0": {
+       "node-addon-api": "7.1.0"
++    },
```
daiyam, line 459: You can't touch the `package.json` since it will affect the others platforms and even Windows.

andreasgrapentin: unfortunately that change is required because the version pulled in by @vscode/l10n-dev causes the build to fail because of an endianness issue. However, that dependency change to package.json should not be a breaking change for any other architecture or platform.

Do you have a suggestion how that change could be limited to the s390x build only?

daiyam: I've just patched the `package_linux_reh.sh` to include patches in `patches/linux/reh/${VSCODE_ARCH}/`.

The build in the CI is a bit cumbersome:
- the first step is to compile the project from TS to JS on x86
- the JS files are then dispatched to each architectures
- the build for an arch is installed the dependencies for the arch

From the current version, it's also how it's done on Windows (built on linux x86) then packaged on Windows. MS does it the same way since at least a year.

By adding a patch in `patches/linux/reh/${VSCODE_ARCH}/`, I'm hoping that the dependencies are correctly installed for your arch.

andreasgrapentin: thank you, I have rebased my changes and moved the part of the patch that changes package.json to that subdirectory. CI build is running now, I'll comment here again when that finishes.

andreasgrapentin:
commit 383d9fd2162860e26491ee439d6db65834f25264 moves the changes to package.json and package-lock.json to a new patch file in patches/linux/reh/s390x/

commit 5b86df781d98144175965f2f2d860c14cb98aede adds the missing slugs in workflows/stable-linux.yml and workflows/insider-linux.yml

CI builds look good on my end.

daiyam:
Can you replace the `vscode-linux-build-agent` containers?

andreasgrapentin:
paths to docker container and sysroot (and shasum) have been updated in fc309f1e6a419af5addf66c1826f714f013dd8e4

daiyam:
The archives are available.

GitMensch:
Thanks guys for your work on this!

(now I only need access to one of those s390x machines with a C compiler on, preferably the IBM one ... )

andreasgrapentin:
> Thanks guys for your work on this!
> 
> (now I only need access to one of those s390x machines with a C compiler on, preferably the IBM one ... )

it's slightly off-topic for this issue, but you can request access to s390 machines with Linux in the LinuxOne cloud free of charge.

PR #2128, opened 2024-11-26
Replace ripgrep binary for riscv64

### LINKED ISSUE #2060: Extension host throws exception "Named export 'rgPath' not found." on RISC-V64 (installed by Open Remote - SSH)
Happens with both VSCodium 1.95.0.24287-insider and 1.94.2.24286. Client is MacOS ARM64, server is RISC-V64 Armbian/Debian Sid Linux.

Log entry - the full log file for 1.95.0.24287-insider is attached [dda161016e115d6fa0fd302d2609be345b578a8b.log](https://github.com/user-attachments/files/17361979/dda161016e115d6fa0fd302d2609be345b578a8b.log):

```text
SyntaxError: Named export 'rgPath' not found. The requested module '@vscode/ripgrep' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@vscode/ripgrep';
```

Originally posted there: https://github.com/jeanp413/open-remote-ssh/issues/173

kxxt:
Fix #2060

Microsoft doesn't respond to my PR https://github.com/microsoft/ripgrep-prebuilt/pull/41 after more than 3 weeks so I decided to host rg binaries for riscv64 in https://github.com/riscv-forks/ripgrep-riscv64-prebuilt .

And a minor fix about the documentation BTW.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,7 @@ The minimal version is limited by the core component Electron, you may want to c
 - [x] GNU/Linux x64 (`deb`, `rpm`, `AppImage`, `snap`, `tar.gz`)
 - [x] GNU/Linux arm64 (`deb`, `rpm`, `snap`, `tar.gz`)
 - [x] GNU/Linux armhf (`deb`, `rpm`, `tar.gz`)
+- [x] GNU/Linux riscv64 (`tar.gz`)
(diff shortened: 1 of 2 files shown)


daiyam:
Good. I was going to ping you about it.

PR #2168, opened 2024-12-20
ci: use jq for JSON parsing, avoid cut for querying version from tag

thican:
Based on recent [job failure](https://github.com/VSCodium/vscodium/actions/runs/12436422943/job/34724736907), I think this should be more reliable; it is not good practice to use text parsing tools (grep, sed, and more) for JSON parsing.
Based on other scripts, `jq` is already used elsewhere so let’s use it there too.

Also use variables TAG and VERSION for readability.

### DIFF
--- a/install_gh.sh
+++ b/install_gh.sh
@@ -4,9 +4,10 @@ set -ex
 
 GH_ARCH="amd64"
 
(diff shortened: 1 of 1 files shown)


daiyam:
Oh, I see now the purpose of this PR... I didn't see the failures.
Weird, I didn't get any notifications.

daiyam:
@thican Thx

thican:
Thanks you, @daiyam, and also all contributors for all this years on this project.

PR #2233, opened 2025-02-22
Add `terminal-suggest` support for the `codium` and `codium-insiders` commands

manuth:
Only recently, Visual Studio Code added experimental support for providing suggestions for auto-completing terminal commands for both the `code` and the `code-insiders` commands by enabling the `terminal.integrated.suggest.enabled` option.

Changes made in this PR extend this behavior by adding the very same auto completion capabilities for the `codium` and the `codium-insiders` commands, too.

### DIFF
--- a/patches/terminal-suggest.patch
+++ b/patches/terminal-suggest.patch
@@ -1,0 +1,84 @@
+diff --git a/extensions/terminal-suggest/src/completions/codium-insiders.ts b/extensions/terminal-suggest/src/completions/codium-insiders.ts
+new file mode 100644
+index 00000000000..a5769b1a575
+--- /dev/null
++++ b/extensions/terminal-suggest/src/completions/codium-insiders.ts
+@@ -0,0 +1,9 @@
++import code from './code';
++
++const codiumInsidersCompletionSpec: Fig.Spec = {
++	...code,
++	name: 'codium-insiders',
++	description: 'VSCodium Insiders',
++};
++
++export default codiumInsidersCompletionSpec;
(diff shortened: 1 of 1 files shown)


daiyam:
Instead of a patch, in `prepare_vscode.sh`, I would add:
```
# terminal-suggest
replace "s|'code'|'codium'|" extensions/terminal-suggest/src/completions/code.ts
replace 's|Visual Studio Code|VSCodium|' extensions/terminal-suggest/src/completions/code.ts
replace 's|code-insiders|codium-insiders|' extensions/terminal-suggest/src/completions/code-insiders.ts
replace 's|Visual Studio Code Insiders|VSCodium Insiders|' extensions/terminal-suggest/src/completions/code-insiders.ts

```

manuth:
Alright then - as recommended, I switched over to using `sed` instead of Patches.
Tysm for the quick review!

daiyam:
@manuth Why the `cp`? You can directly modify the existing files.

manuth:
@daiyam that's in order to leave the original `code` and `code-insiders` autocompletes intact

manuth:
tbh, just because it's easily doable
I guess there is also a high likelyhood that people - much like me - have both `code` or `code-insiders` and `codium` or `codium-insiders` CLIs available on their systems so it's quite handy to have auto-complete for all CLIs in the terminal.

daiyam:
Sorry for the delay

> @daiyam that's in order to leave the original `code` and `code-insiders` autocompletes intact

Ok to have both.
In that case can you add back the patch version. (it was removed from the history since you have done a force-push)

manuth:
No worries - thanks for coming back at me! 😄
Did so

manuth:
Awesome, thank you so much!

PR #2253, opened 2025-03-11
Update README.md with new insider installation for macOS

AM-I-Human:
I got an error while installing the insider version, here the fix to 
brew install --cask vscodium@insiders


daiyam:
Thx

PR #2264, opened 2025-03-14
Fix error with qemu and ppc64le segfaults on build reh

### LINKED ISSUE #2261: VSCodium REH 1.98.1.25070 for PPC64LE build failed with SegFault
Looks like the build for the latest release of VSCodium REH failed on PowerPC with a segfault. Has this happened before? Should I look into it? Could we try triggering that job again?

https://github.com/VSCodium/vscodium/actions/runs/13795293397/job/38586077050

lex-ibm:
Fixes #2261

Also adds an option to disable this QEMU step by `export DISABLE_QEMU=true`

### DIFF
--- a/patches/linux/yarn-dependencies.patch
+++ b/patches/linux/yarn-dependencies.patch
@@ -1,12 +1,17 @@
 diff --git a/build/npm/postinstall.js b/build/npm/postinstall.js
-index 88e3c9e..4cad4f1 100644
+index c1f22aa5002..a6fb2728b1f 100644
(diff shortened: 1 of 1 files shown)


daiyam:
Thanks for the catch. It might be related to the difficulties I had to build reh for 1.98...

lex-ibm:
Maybe, I was trying to build using docker-in-docker and qemu would just mess things up for me, so I disabled it and build passed. The image used in that script is also no longer the recommended way.
Anyhow, thanks for accepting the change!

PR #2265, opened 2025-03-14
Propose devcontainer environment and tasks

lex-ibm:
Bare-bones devcontainer and a few simple tasks to make it easier when debugging locally.

I was going to make the Containerfile more Power focused but I realized a lot could be left generic and changed using a `.env`. 

The tasks are just enough to build REH:
- `Download latest stable-linux vscode artifact`: Downloads and extracts the latest vscode.tar.gz from stable-linux master (Hopefully that is available at the time someone wants to run this task and it hasn't expired)
- `Extract vscode.tar.gz`: If the `vscode.tar.gz` is already present, just extract it and change the ownership
- `Build Linux REH`: Run the `./build/linux/package_reh.sh`

Hopefully this helps others.

### DIFF
--- a/.devcontainer/Containerfile
+++ b/.devcontainer/Containerfile
@@ -10,6 +10,7 @@ RUN apt update && \
         libxkbfile-dev \
         libsecret-1-dev \
         qemu-user \
+        shellcheck \
         && \
     apt clean && \
     rm -rf /var/lib/apt/lists/* && \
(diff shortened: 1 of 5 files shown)


lex-ibm:
@daiyam as mentioned in #2261.

Dev environments tend to be a bit too opinionated sometimes 😅

PR #2273, opened 2025-03-18
fix: use electron v34.3.2

daiyam:


tmm1:
fyi electron <34.3.3 doesn't run on rhel8 https://github.com/electron/electron/issues/45884

PR #2348, opened 2025-04-27
fix(reh): use correct ripgrep.sh path

darkyzhou:
This PR fixes the incorrect `ripgrep.sh` path specified in `package_reh.sh`, causing `rg` binary of REH server not being replaced with native ones.


daiyam:
Oh! Thanks for catch the error!

PR #2350, opened 2025-04-28
fix: migrate from el9 to el8

daiyam:


lex-ibm:
Hmmmm I think I might need to build that version of electron

PR #2386, opened 2025-05-22
snap: Use native wayland client when possible

3v1n0:
Electron supports wayland natively since version 12, but when running code from the snap it's not possible to use because of some issues.

Handle them here so that a native client can be used when in wayland (unless `DISABLE_WAYLAND` env variable is used), making code to have proper rendering in HiDPI screens.

The wayland behavior can now controlled using `snap set codium wayland-native=false` or `snap set codium wayland-native=true` (by default it's still disabled).

---

This is a similar to https://github.com/microsoft/vscode/pull/199052 but without touching the actual code and it is as it was originally done for it, although it didn't get ever in.


daiyam:
@3v1n0 Thanks for the PR but could you rebase it on the `insider` branch?

3v1n0:
Done!

PR #2407, opened 2025-06-26
Update extensions-compatibility.md

r3m0t:
Python extension license is MIT. Pylance license is closed source as shown here

https://marketplace.visualstudio.com/items/ms-python.vscode-pylance/license


daiyam:
Being MIT doesn't mean that it's fully compatible with VSCodium...

r3m0t:
I agree but I'm not aware of any issue. I believe this was written in confusion because it auto installs the Pylance extension which is licensed only for GitHub and MS IDEs.

BilgiPasa:
I think you cannot install Microsoft's Pylance extension in VSCodium but you can install Microsoft's Python extension.

In VSCodium, when I install Microsoft's Python extension, it also auto installs Python Debugger and Python Environments extensions.

In VSCodium, because I cannot install Pylance, I use BasedPyright instead (which I think that it is better than Pylance).

When I tried to use Microsoft's Python extension with BasedPyright, I think it works very well.

BilgiPasa:
> Being MIT doesn't mean that it's fully compatible with VSCodium...

But like you said, it does not mean that it is fully compatible with VSCodium. I did not do any serious, in depth testing. So, there might be a problem that I don't know of.

PR #2543, opened 2025-10-09
fix(snap): fix snapcraft build files

daiyam:


3v1n0:
Isn't this basically reverting https://github.com/VSCodium/vscodium/commit/19811fd8245a8607f3ff06a1c2e7929083b38d4b ? Let me know if I can help here, but definitely this should break the wayland integration, no?

daiyam:
No, it's working fine with Wayland.

There were issue in the PR. Can find the original issue of June but I had to revert the `1.101` release.
Here the errors some users and I were getting:
```
[1692752:0922/144000.258875:ERROR:gpu_process_host.cc(953)] GPU process exited unexpectedly: exit_code=139
[1692752:0922/144000.258897:WARNING:gpu_process_host.cc(1389)] The GPU process has crashed 9 time(s)
```

I think it was due to the extra `.so` files.

3v1n0:
I will test this again, but one of the reasons for having different libraries was to get AMD gpus to work

PR #2605, opened 2025-11-21
Update macOS runners in GitHub Actions workflow to macOS 15

pheiduck:
https://github.com/actions/runner-images/issues/13046

macOS 13 will be fully unsupported by December 4th for GitHub and ADO


pheiduck:
@daiyam insider runner is also updated to macos-15

pheiduck:
Adopted already.

PR #2675, opened 2026-01-23
Update Node.js version in build documentation

andrewhertog:
Slight discrepency in the docs, that needs updating. I ran into an issue with node 20, due to the increased use of ts, i was getting unknown filetype error, and then i found that the nvmrc had been updated since i last built this.

### DIFF
--- a/docs/howto-build.md
+++ b/docs/howto-build.md
@@ -17,7 +17,7 @@
 
 ## <a id="dependencies"></a>Dependencies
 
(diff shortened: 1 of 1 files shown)


daiyam:
I've added the reference to the `.nvmrc` file since it can change every month. What do you think?

andrewhertog:
Yup, makes sense to me.

daiyam:
Thx

PR #2723, opened 2026-03-02
docs: add Windows troubleshooting section (GPO, context menu, Defender)

### LINKED ISSUE #2714: VSCodium ignoring GPOs
**Describe the bug**
We are looking into replacing VSC with VSCodium in our Enterprise. I have tested MS VSCode GPOs that are delivered with VSCode and they work fine. I have now installed VSCodium, applied the same policy from gpedit.msc => computer configuration => administrative templates =>VSCodium and checked them in regedit. they appeared here: HKLM\SOFTWARE\Policies\Microsoft\VSCodium. They look exactly the same like the VSC ones so I thought we are all set. But they are completely ignored.

This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.

I haven't found similar problem.

**To Reproduce**
Steps to reproduce the behavior locally, outside of the domain to make it as simple as possible:
- copy admx/adml templates (2) to correct folders: admx goes to C:\Windows\PolicyDefinitions\ while adml goes to en-us subfolder.
- open gpedit.msc, find vscodium group, select for example UpdateMode and set it to none.
- you dont need to restart if you test all locally (outside of the domain). Just launch VSCodium and go preferences, find update settings and check if you have Update: Mode set to none with "managed by organization" in the brackets.
- it works exactly like this when I use the same version of VSC.

**Expected behavior**
It should not ignore the GPOs.

**Screenshots**
Here are 2 apps opened at the same time , using exactly the same policies, set from gpedit,msc
- 1st is VSCode 1.109.5 x64
- 2nd is VSCodium 1.109.51242 x64

Image

Registry:

Image

 - OS: Windows 11 PRO 25H2
 - x64
 - Microsoft Windows [Version 10.0.26200.7840]
 - App has been installed from ZIP and via installer.
 - No virtualization.

**Additional context**
I have tested this in our domain and on my Private machine. The results are the same.
- Company domain endpoint: I have set all registry manually, 1:1 like  VSC. Doesnt work.
- Private machine. I have set desired GPO via gpedit.msc to avoid typos and similar stupid mistakes. Doesn't work either.

bhaskarvilles:
- Document correct registry path for VSCodium GPOs: HKLM\SOFTWARE\Policies\VSCodium\VSCodium\<PolicyName>
- Add Windows 11 context menu workaround (Shift+Right-click)
- Add Windows Defender false-positive guidance
- Add explanatory comment to fix-policies.patch Fixes: #2714 (documentation)

### DIFF
--- a/docs/patches.md
+++ b/docs/patches.md
@@ -1,0 +1,43 @@
+# Patches
+
+Documentation for VSCodium patches applied on top of VS Code.
+
+---
+
+## fix-policies
+
+**Replace `@vscode/policy-watcher` with `@vscodium/policy-watcher`**
+
+VS Code uses `@vscode/policy-watcher` to enforce Group Policy Objects (GPOs) on
+Windows. That package reads from:
+
+```
+HKLM\SOFTWARE\Policies\Microsoft\<productName>
+```
+
+VSCodium forks this into `@vscodium/policy-watcher`, which takes a separate
+`vendorName` argument. The `createWatcher()` call becomes:
+
+```ts
+createWatcher('VSCodium', this.productName, ...)
+```
+
+Because VSCodium sets `product.nameLong = 'VSCodium'` (via `prepare_vscode.sh`),
+`this.productName` resolves to `'VSCodium'` at runtime. Therefore, the final
+Windows registry key that VSCodium reads policies from is:
+
+```
+HKLM\SOFTWARE\Policies\VSCodium\VSCodium\<PolicyName>
+```
+
+(or `HKCU\SOFTWARE\Policies\VSCodium\VSCodium\<PolicyName>` for per-user policies)
+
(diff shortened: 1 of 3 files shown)


daiyam:
Can you move the documentation added in `fix-policies.patch` to a new file named `patches.md` and under the section `fix-policies`?
If left in `fix-policies.patch`, it will be overwritten when in the next fix.
The `patches.md` file is on my TODO list...

bhaskarvilles:
@daiyam , Done! I've created 

docs/patches.md
 with a ## fix-policies section containing all the context from the patch header. The patch file now just has a single pointer comment:
 
 `# Documentation: docs/patches.md#fix-policies`

daiyam:
Thanks

PR #2766, opened 2026-03-27
Fix DEB issue with PowerPC64 Little Endian

lex-ibm:
PowerPC64 Little Endian is named `ppc64el` in Debian and `ppc64le` in EL/Fedora


lex-ibm:
Also @daiyam, could we enable building and releasing RPMs and DEBs for PPC64le?

PR #2778, opened 2026-04-04
Fix issue with RPMs and DEBs when packaging for PowerPC64 LE

lex-ibm:
This would be a temporary fix, but I would really like to centralize where the arch-specific configurations live

### DIFF
--- a/.github/workflows/insider-linux.yml
+++ b/.github/workflows/insider-linux.yml
@@ -231,6 +231,7 @@ jobs:
         if: env.DISABLED != 'yes' && env.SHOULD_BUILD == 'yes'
 
       - name: Build
+        id: build
         env:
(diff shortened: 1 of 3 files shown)


lex-ibm:
@daiyam any thoughts?

daiyam:
It needs to use the variables defined in `package_bin.sh`.

lex-ibm:
Yes, it does, but wouldn't sourcing the script add more build time?

daiyam:
> Yes, it does, but wouldn't sourcing the script add more build time?

`package_bin.sh` needs to export those variables and the workflow needs to pass them back to `prepare_assets.sh`

lex-ibm:
@daiyam something like this?

daiyam:
https://github.com/VSCodium/vscodium/releases/tag/1.116.02821

PR #2790, opened 2026-04-16
enable having space-separated args for terminal.external.linuxExec

timkenhan:
Currently, the setting terminal.external.linuxExec can only be processed as a single string. This makes it impossible to have args in the exec command. Workaround exists in the form of creating a script with the desired command and its arguments, but this is far from clean.
It is also nowhere near efficient, especially in the case where the user only wants to add an option or two to modify their terminal behavior. It is especially painstaking in properly having external terminal setting when running it as Flatpak application (where one would need to wrap the command in flatpak-spawn).


patches/linux-extterm-args.patch
```diff
@@ -0,0 +1,20 @@
...
+--- a/src/vs/platform/externalTerminal/node/externalTerminalService.ts
++++ b/src/vs/platform/externalTerminal/node/externalTerminalService.ts
+@@ -325,9 +325,12 @@ export class LinuxExternalTerminalService extends ExternalTerminalService implem
+ 		return new Promise<void>((c, e) => {
+ 			execPromise.then(exec => {
+ 				const env = getSanitizedEnvironment(process);
+-				const basename = path.basename(exec).toLowerCase();
+-				const args = basename === 'ghostty' && cwd ? [`--working-directory=${cwd}`] : [];
+-				const child = spawner.spawn(exec, args, { cwd, env });
++				const args: string[] = exec.split(/(?<!\\)\s/);
```
daiyam: Please improve the split, currently it would also split strings like `"hello world"`

daiyam:
Is there any related issues or PR on vscode?

timkenhan:
> Is there any related issues or PR on vscode?

That would be [this](https://github.com/microsoft/vscode/pull/308965)

> currently it would also split strings like "hello world"

What would be the expected behaviour for this? Would it consider the space escaped? If so, lemme see if I can expand the regex.

daiyam:
> What would be the expected behaviour for this? Would it consider the space escaped? If so, lemme see if I can expand the regex.

I would call this a standard issue when parsing/splitting the command line since we can have a command like `print "Hello world"` and want to have `{cmd: 'print', args: ['"Hello world"']}` but not `{cmd: 'print', args: ['"Hello', 'world"']}`...

timkenhan:
@daiyam I've found a much simpler solution of having the `{ shell: true }` option in the `spawn` method. This should address your concern as well.

daiyam:
Not sure, if it should be security issue (https://www.sourcery.ai/vulnerabilities/javascript-lang-security-detect-child-process) or not.
For local dev, it shouldn't since the user have already "full" access.
It effect will need to evaluated for the web version (todo: try https://github.com/aadityapurani/NodeJS-Red-Team-Cheat-Sheet)

timkenhan:
@daiyam I just checked and terminal is not available on the web version. For the Linux desktop version, I don't think this should be a concern either, considering the usecase.

timkenhan:
@daiyam do you think we can proceed?

daiyam:
@timkenhan Can you disable the workflows in your fork? I'm getting some alerts from it about the automated workflows (https://github.com/timkenhan/vscodium/actions/runs/24937878249)

timkenhan:
@daiyam I've disabled the Actions on the repo settings. Hopefully this should do. I'm not too sure myself as this seems to have came with the fork.

PR #2817, opened 2026-05-01
Fixed arguments on appimage

redsigma:
When doing commands such as below, it opens an instance of codium instead of listing active extensions in cli or installing extensions. 
```
vscodium.AppImage --list-extensions

vscodium.AppImage --install-extension

```

This PR fixes that by using a simpler approach rather than adjusting for each possible command. The following syntax will properly work and installing extensions are also properly working even with a custom --extensions-dir
```
vscodium.AppImage -- --list-extensions
vscodium.AppImage -- --install-extension
```

For running the app normally  you can still use the following

```
vscodium.AppImage

vscodium.AppImage --user-data-dir /home/user/mycustomdata  --extensions-dir /home/user/mycustomextension
```


daiyam:
@redsigma Can you target the `insider` branch? Thx

redsigma:
Moved here https://github.com/VSCodium/vscodium/pull/2818

PR #2821, opened 2026-05-02
feat: remove update version since we use gh releases

daiyam:


daiyam:
No sure what got in me, but it's completely wrong...

But, yeah, I have to remove the use of that `versions` repo but it will for next year.

PR #2831, opened 2026-05-11
fix: ensure build stops if prepare_vscode fails

Iheb-AG:
Add error handling for preparation step in build script

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -9,6 +9,10 @@ if [[ "${SHOULD_BUILD}" == "yes" ]]; then
   echo "MS_COMMIT=\"${MS_COMMIT}\""
 
   . prepare_vscode.sh
+  if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


daiyam:
It already taken care with `set -ex`.

Which error does it fail to catch?

Iheb-AG:
You're right that set -e covers most failures. However, during local testing, I found that if prepare_vscode.sh is sourced, certain shell environments allowed build.sh to continue despite internal errors. This explicit check ensures the build strictly halts the moment preparation fails, preventing the system from wasting resources on a broken build process.

daiyam:
> certain shell environments

Can you be more explicit?

Iheb-AG:
I am using fish as my shell. When running the build, I noticed that if an error happens inside the sourced prepare_vscode.sh, the main build.sh doesn't always stop—it just keeps going.

PR #2858, opened 2026-05-20
Add official serviceUrl/itemUrl

olfek:
Thanks to:
https://zhuanlan.zhihu.com/p/276000982


daiyam:
They were removed from the docs (https://github.com/VSCodium/vscodium/pull/2113) because Microsoft told me that it was breaking their terms of use.
So I won't add them back. Sorry.

PR #2903, opened 2026-06-29
feat(1.126): update patches

daiyam:


daiyam:
@kxxt I can't figure out why the REH riscv64 can't be built. Can you take a look? Thx

kxxt:
> @kxxt I can't figure out why the REH riscv64 can't be built. Can you take a look? Thx

Thanks for the ping! I suspect that the follow log lines are related to the failure. I am implementing a fix now.

```
npm warn Unknown env config "arch". This will stop working in the next major version of npm. See `npm help npmrc` for supported config options.
npm warn Unknown project config "build_from_source". This will stop working in the next major version of npm. See `npm help npmrc` for supported config options.
npm warn Unknown project config "timeout". This will stop working in the next major version of npm. See `npm help npmrc` for supported config options.
```

It appears the npm version used here no longer supports `npm_config_arch`.

kxxt:
> @kxxt I can't figure out why the REH riscv64 can't be built. Can you take a look? Thx

I think this is caused by https://github.com/VSCodium/vscode-linux-build-agent/commit/a7a79899ad4ea4c7e3926c4600c07cb0a741c4a7 . The `focal-devtoolset-riscv64` should use a riscv64 node.js binary. However, it switched to use a x64 node binary in that commit.

I tried to switch back to the old unofficial-builds binary but the 24.15.0 binary there requires a higher glibc version.
I am checking if other v24.x riscv64 binaries have such high glibc version requirement.
If there's none that could work, I will build one myself.

kxxt:
> I am checking if other v24.x riscv64 binaries have such high glibc version requirement.
> If there's none that could work, I will build one myself.

Unfortunately there's no binary that could work in `focal-devtoolset-riscv64`. The new binaries are targeting glibc 2.38 while  `focal-devtoolset-riscv64`  uses glibc 2.31.

I will implement a workflow to publish a node v24.x binary targeting glibc 2.33 for riscv64 in https://github.com/riscv-forks/node-riscv when I have time. 
After that, the path to move forward is dropping `focal-devtoolset-riscv64`  and add `jammy-devtoolset-riscv64` as a replacement, which ships glibc 2.35.

daiyam:
I did switch to `https://deb.nodesource.com/setup_24.x` since `focal-riscv64` was already using it.
Yeah, nodesource only supports amd64 and arm64...

A higher glibc doesn't bother me. There is already a `jammy-devtoolset-riscv64`, I tried it to see if it was better (same previous issue).
You can upgrade the package if you want. Not sure if there a lot of people running a 2020 system on RiscV64...

kxxt:
> You can upgrade the package if you want. Not sure if there a lot of people running a 2020 system on RiscV64...

I don't think there will be anyone caring about using glibc 2.31 on riscv64 now.
So I will upgrade it to `jammy` with glibc 2.35.

Here is the PR: https://github.com/VSCodium/vscode-linux-build-agent/pull/43

> I tried it to see if it was better (same previous issue).

But still, the node binary on unofficial-builds requires glibc 2.38, which explains why it failed.
IMO glibc 2.38 requirement is too tough.  So I implemented a CD pipeline that builds node.js against my Chromium debian riscv64 sysroot, which supports glibc 2.33.

After the vscode-linux-build-agent PR is merged, the next step is to replace `focal-devtoolset-riscv64` with `jammy-devtoolset-riscv64` in this repo and also replace the node.js binary download link for riscv64 in this repo.

PR #2905, opened 2026-07-01
Address EditorConfig violations

BaumiCoder:
This Pull Request is part of an academic research project. More preciously it is for my master's thesis, more background information [here](https://codeberg.org/BaumiCoder/Masterthesis).

## Definition: EditorConfig violations

The goal of this Pull Request is to address all [EditorConfig](https://editorconfig.org) violations in the project. With EditorConfig you have set for example `insert_final_newline = true` for most files, but there were files without final newlines. I call this an EditorConfig violation as a file content violates the respective EditorConfig property. They can have different kind of impacts. For example trailing spaces, which get remove with the next save, may make the next diff / Pull Request to this file less readable. However, missing final newline may lead to problems when reading the file with some application.

## How to address them?

To check for such violations and fix them I developed the CLI tool [ecformat](https://codeberg.org/BaumiCoder/ecformat). However, there are different ways to address the violations:

1. **Adjust EditorConfig:** The EditorConfig properties for a file can be wrong. For example two trailing spaces in Markdown represent a line break. Therefore, removing them would break the layout. The correct way to address that is to set `trim_trailing_whitespace = false`  for them, which avoids automatic removal of such spaces.
2. **Fix file content:** Another way is to fix the content of the file. Which means for example for `insert_final_newline = true` to insert the final newlines. To do so `ecformat fix` can be helpful.
3. **Ignore files:** For some files it is not meaningful to consider the EditorConfig violations, which `ecformat check` finds. I added a `.ecformat_ignore` file for this. It has the common ignore syntax such as a `.gitignore`. To use such ignore files, provide the name of them with an CLI switch, `ecformat check --ignore-file .ecformat_ignore`. (The `.gitignore` files are automatically considered inside a git repository.) There are different reasons to ignore files:
   1. As EditorConfig is only for text files, considering binary files does not make sense.
   2. EditorConfig supports to set the property on the file level. However, there can be files with intentional differences inside. Such files need to be ignored fully by `ecformat` as it currently (version 0.2.0) has no support to ignore only a single property for some files.

## Changes in this Pull Request

For this Pull Request, I addressed all violations in the project in the way, which looks the most correct one to me. (I focused on the Properties of the current [EditorConfig specification](https://spec.editorconfig.org/#supported-pairs) 0.17.2.) If I fixed file contents, I manually reviewed all of them, making adjustments in the `ecformat fix` changes were necessary.

If I picked the wrong way to address the violations for some files, please let me know with a respective Review comment.
Then I will address it in the correct way. I add Review comments myself for special parts, which are especially worthy of discussion. To support the Review of this Pull Request, I tried to group the changes in meaningful commits with respective commit messages. Therefore, may take a look at the commit history.

## Avoid EditorConfig violations in the future

Such EditorConfig violations can creep in again. Possible reasons can be that the editor of some contributors has no EditorConfig support, or it would be required to install a respective plugin for the support.

If you want to avoid them in the future, I would suggest integrating a respective utility. For example as part of the CI.
One possibility is to use `ecformat` with `ecformat check --ignore-file .ecformat_ignore`. Another one is [editorconfig-checker](https://editorconfig-checker.github.io), which is a more sophisticated tool without the possibility to fix violations.
It provides more configuration options, such as ignoring violations in single lines with marker comments.

Let me know if you are interested using one of them. I could add one in this Pull Request or a follow-up Pull Request. If you not consider to use `ecformat` in your project, the `.ecformat_ignore` file is maybe not worth to merge into the master branch. I can remove it if desired.


build/windows/msi/i18n/vscodium.de-de.wxl
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
BaumiCoder, line 1: In the `.wxl` files and several other XML-lke files I only remove the UTF-8 BOM (as `charset` is `utf-8` and not `utf-8 bom`). The ru-ru file was already without the BOM and for XML files the BOM is optional by [definition](https://www.opentag.com/xfaq_enc.htm#enc_bom).

.editorconfig
```diff
@@ -17,9 +17,22 @@ indent_size = 2
...
 
-[*.patch]
+[*.md]
+# Two trailing spaces in Markdown are a line break
+trim_trailing_whitespace = false
+# Markdown has heavily context specific indentations.
+# For example, the indentation in lists depend on the length of the list marker
+# (such as '- ' or '100. '), at least in the GitHub Markdown flavor:
+# https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#nested-lists
+# Therefore, pressing tab should insert a single space to encourage situative accurate indentations.
```
BaumiCoder, line 27: It is maybe a matter of taste if inserting only one space with a tab is a nice solution for this or not. Another one would be to keep the default `indent_size = 2` for Markdown files and ignore them when checking for EditorConfig violations.

.editorconfig
```diff
@@ -17,9 +17,22 @@ indent_size = 2
...
+indent_size = 1
+
+[*.patch{,.no,.yet}]
 trim_trailing_whitespace = false
 insert_final_newline = false
+# There are files with different kinds of indentations
```
BaumiCoder, line 33: Another way to address that to keep some indentation properties set, which are suitable for most of the patches and ignoring these files when checking for EditorConfig violations.

LICENSE
```diff
@@ -20,4 +20,4 @@ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
 AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
```
BaumiCoder, line 23: Here I also added only a final newline. I don't know why GitHub does not show the "⛔" symbol as in the other files, that here was no final newline before.

daiyam:
@BaumiCoder It's great that you made `ecformat` but it needs to be available outside `cargo`. (`nix`, `npm`, `pipx`, ...) It would be easier to integrate to current workflows (see `zizmor`).

It would be great to be able to add to `.lintstagedrc.yml` (https://github.com/lint-staged/lint-staged)

daiyam:
I'm using it on https://github.com/jeanp413/open-remote-ssh...

It shouldn't change binary files like `.png` or `.gif`

Edit: I will open some issues on you repo ;)

BaumiCoder:
> @BaumiCoder It's great that you made `ecformat` but it needs to be available outside `cargo`. (`nix`, `npm`, `pipx`, ...) It would be easier to integrate to current workflows (see `zizmor`).
> 
> It would be great to be able to add to `.lintstagedrc.yml` (https://github.com/lint-staged/lint-staged)

I plan to make it available in further ways. See for example [issue 13](https://codeberg.org/BaumiCoder/ecformat/issues/13). Nice to hear that this would be really helpful.

daiyam:
Sorry, I will look at your PR after I've released a new version of VSCodium.

BaumiCoder:
> Sorry, I will look at your PR after I've released a new version of VSCodium.

@daiyam The release was several weeks ago. May you have time for this PR in the next few weeks?

daiyam:
I will.

daiyam:
Firstly, I find it weird that requires `--ignore-file .ecformat_ignore`. `.ecformat_ignore` or a `.ecignore` should be automatically included.

I've moved XML to `indent_style = tab` but the `fix` is breaking the indentations so all the lines are only indented by 2 spaces...

BaumiCoder:
> Firstly, I find it weird that requires `--ignore-file .ecformat_ignore`. `.ecformat_ignore` or a `.ecignore` should be automatically included.

Fair point, I created an [issue](https://codeberg.org/BaumiCoder/ecformat/issues/66) to add this to ecformat.

> I've moved XML to `indent_style = tab` but the `fix` is breaking the indentations so all the lines are only indented by 2 spaces...

The `indent_size` for the XML files was still 2, while only the `tab_width` was changed to 4. This implies that one indentation level has size 2, but one tab character is too big with its width of 4. Therefore, 2 spaces = 1 indentation level, 1 tab = 2 indentation levels, 1 tab + 2 spaces = 3 indentation levels, 2 tabs = 4 indentation levels and so on. This is how such differences between `tab_width` and `indent_size` is [specified](https://spec.editorconfig.org/#indentation-non-normative). If I understand you correctly, you want 1 tab = 1 indentation level, while 1 tab having a width of 4. I adjusted the EditorConfig accordingly with 623bd81eea4693eadf7b46d95a41b48967f59b2c. (By the way, `tab_width` defaults to the value of `indent_size`. If `tab_width` never differ from `indent_size` you do not have to set it in your whole `.editorconfig` file.)

I also adjusted the XML files to match the new configuration (see 980fae6abb1ac669a6d5be8eb030d2d037b723fa). For such bigger indentation changes, ecformats capabilities are limited as it has no syntax parsing and do not know the old indentation settings. Therefore, I replaced the leading couples of spaces with one tab character each.

daiyam:
Yes, adding the `indent_size = 4` make it work correctly.
I find it weird to have a `indent_size` different from the `tab_width`...

I usually have the following root config so it has never occurred to me:
```
[*]
indent_style = tab
tab_width = 4
```

daiyam:
LGTM

@BaumiCoder Do you have anything else to add?

BaumiCoder:
Looks good to me. Thanks for your time.

Only one thing, about usage of `indent_size` and `tab_width`. You wrote:

> Yes, adding the `indent_size = 4` make it work correctly. I find it weird to have a `indent_size` different from the `tab_width`...

If you never want to have a difference between `indent_size` and `tab_width`, you never have to set `tab_width` as it defaults to `indent_size`. The other way round is also possible by setting `indent_size = tab`, see [EditorConfig documentation](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_size). (I have no idea when ones want to have different values between `indent_size` and `tab_width`.)

daiyam:
> If you never want to have a difference between `indent_size` and `tab_width`, you never have to set `tab_width` as it defaults to `indent_size`. The other way round is also possible by setting `indent_size = tab`, see [EditorConfig documentation](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_size). (I have no idea when ones want to have different values between `indent_size` and `tab_width`.)

Yes, it appears more logical. Thx

daiyam:
@BaumiCoder It would be great if there was a faster way to install `ecformat` than with `cargo`.

BaumiCoder:
> @BaumiCoder It would be great if there was a faster way to install `ecformat` than with `cargo`.

Currently, there would be also the possibility to download the binary for several common platforms from [here](https://codeberg.org/BaumiCoder/-/packages/generic/ecformat_bin). More handy methods will come in later versions, see [issues](https://codeberg.org/BaumiCoder/ecformat/issues?labels=588082).

PR #2912, opened 2026-07-05
fix(loong64): update electron

daiyam:


lex-ibm:
Should I build this version for PowerPC64 LE?

daiyam:
@lex-ibm Ya, if you want. But a fix for REH would be more urgent. I'm planning to release the `1.126` tomorrow.

PR #2921, opened 2026-07-09
fix(riscv64): fix REH builds

kxxt:
This PR fixes the REH build failures that first occurred in https://github.com/VSCodium/vscodium/pull/2903.

It increases the glibc version requirement for riscv64 REH builds to 2.35, because ubuntu jammy's cross toolchain uses glibc 2.35 (while the new node.js binary used in this PR supports 2.33).

However, the desktop application builds should still support glibc 2.33, because it is still using `focal-riscv64` whose  cross toolchain uses glibc 2.31 (while the electron riscv64 binary supports 2.33)

The first commit switches to the jammy image. The second one replaces the node.js binary from unofficial-builds with my cross-compiled binary.

### DIFF
--- a/build/linux/package_reh.sh
+++ b/build/linux/package_reh.sh
@@ -39,7 +39,7 @@ elif [[ "${VSCODE_ARCH}" == "ppc64le" ]]; then
   export VSCODE_SYSROOT_REPOSITORY='VSCodium/vscode-linux-build-agent'
   export VSCODE_SYSROOT_VERSION='20260706'
 elif [[ "${VSCODE_ARCH}" == "riscv64" ]]; then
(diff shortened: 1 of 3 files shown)


kxxt:
The compile step is failing with

```
++ echo 'Not found: src/vs/platform/agentHost/node/codex'
Not found: src/vs/platform/agentHost/node/codex
```

daiyam:
Merge with the `master` branch.
You are still using a older/broken branch https://github.com/VSCodium/vscodium/actions/runs/29022393417/job/86132818324?pr=2921#step:8:23

kxxt:
> Merge with the `master` branch. You are still using a older/broken branch https://github.com/VSCodium/vscodium/actions/runs/29022393417/job/86132818324?pr=2921#step:8:23

Thanks! I have rebased it on top of master branch. Initially I used the insider branch.

daiyam:
The REH archives are available

PR #2961, opened 2026-08-10
Improve desktop entry

### LINKED ISSUE #2960: AppImage does not run after installation
I installed the Codium AppImage file via AppManager but the application does not run.
https://github.com/kem-a/AppManager

I need to change the Command-line arguments to run the application:

Image

Image

This change has already been made in other electron applications. (like Obsidian)

xmha97:
This issue fix https://github.com/VSCodium/vscodium/issues/2960


daiyam:
Explain why it should be applied to all packages/distros and not only to AppImage installed via that AppManager?

Also, do not upvote your own issue or PR. Thx

xmha97:
I thought `--no-sandbox` was required for all distributions.

daiyam:
> I thought `--no-sandbox` was required for all distributions.

In regular `.deb` or `.rpm`, there is no sandbox. As `.appimage` which aren't sandboxed by default.

xmha97:
Thank you for your explanation.

PR #2967, opened 2026-08-17
Proposal: Change how patches are managed and applied

lex-ibm:
**This PR has significant contributions made by AI.**

My idea was to use the same mechanism as Electron does to apply their patches, mostly because I became tired of having to figure out what was mine and what was other people's work.

This is still a WIP, but feedback is always welcomed.


daiyam:
Ok, I've tested it.

Here, what I like:

- The header in the patches. It's clearer about what they do.
- To see the patch note in the commit message when edit a patch. It's nice.

What need to be improved:

- It would be easier to track a patch by its filename than its subject (in the header)
  `Applying: remove welcome onboarding screen`, not clear which file is being applied, need to do search to find the file.
- The support for json patches (like `52-ext-copilot-remove-it.json`) need to be improved. Or to be skippable.
- In the incoming, the patch code was mixed with old VSCode code. I had to check the patch to verify what it was doing. `.rej` is a bit easier here.
- The `rebase.sh` is a bit of a black box. I wanted to stop after few patches. `./dev/export.sh` told me to `Finish or abort the rebase/am in vscode` so I aborted it. Then, a second `./dev/export.sh` didn't produce any new patches...

lex-ibm:
@daiyam I believe I've addressed your feedback in the latest push, but I just noticed the patches might be out of sync

daiyam:
@lex-ibm Don't update to the last `insider` so it can be a test case for an update (from `126` to `132`)

lex-ibm:
Oh, I had just done that
Let me revert

lex-ibm:
@daiyam back to before rebasing to 135

Issue #2049, opened 2024-10-07
Remote Extension Host: "The requested module 'node:module' does not provide an export named 'register'" on Linux


MicrosoftEggshell:
**Describe the bug**
When attempting to connect to a remote host, I get `Error: Couldn't install vscode server on remote server, install script returned non-zero exit status`. 

This happened specifically using [open-remote-ssh](https://github.com/jeanp413/open-remote-ssh), which [uses](https://github.com/jeanp413/open-remote-ssh/blob/master/src/serverSetup.ts#L38) VSCodium's REH build. When manually downloading and attempting to execute `bin/codium_server` on the server, I get the error `"The requested module 'node:module' does not provide an export named 'register'"`.

This only happens on the latest release build ([1.94.0.24281](https://github.com/VSCodium/vscodium/releases/tag/1.94.0.24281)), not the [previous](https://github.com/VSCodium/vscodium/releases/tag/1.93.1.24256) one. The Windows version of REH also works just fine.

**IMPORTANT** (for those reading who just want to use remote): A quick and simple workaround is as follows:
1. Attempt to connect to the remote server (this installs vscodium-server)
2. Stop all running instances of the bundled node on the server (eg. by running `pkill -f ~/.vscodium-server/bin/*/node` as the ssh user, or `pkill -f .vscodium-server/bin/*/node` as any user)
3. Replace `~/.vscodium-server/bin/<commit>/node` on the server with a newer [version](https://nodejs.org/en/download/prebuilt-binaries) of the node binary, **v18.20.4** works. Once again, `~` here means the home directory of the user you're using to connect with SSH.

**Please confirm that this problem is VSCodium-specific**
- [x] This bug doesn't happen if I use Microsoft's Visual Studio Code. It only happens in VSCodium.
  - The official stable server (downloaded by the Microsoft's [Remote - SSH](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-ssh) extension) uses node v20.16.0

**Please confirm that the issue/resolution isn't already documented**
- [x] I checked [the Docs page](https://github.com/VSCodium/vscodium/blob/master/docs/index.md) and [the Troubleshooting page](https://github.com/VSCodium/vscodium/blob/master/docs/troubleshooting.md) and my issue is not mentioned there.

**To Reproduce**

Download and extract `vscodium-reh-linux-x64-1.94.0.24281.tar.gz` on a Linux machine, and run `bin/codium_server`

**Expected behavior**
[...]

PR #2978, opened 2026-08-27
Fix inverted `enableASARSupport()` guard

0zitro:
Subject :: <https://github.com/VSCodium/vscodium/pull/2973/changes#diff-a7a7cf28296f4f756ff70fe19553b7d3661fe16379a7b64c334fe05532116b6bR10-R14>

```patch
 function enableASARSupport(): void {
-     if (!process.env['ELECTRON_RUN_AS_NODE'] && !process.versions['electron']) {
+     if (Module.register as unknown && (process.env['ELECTRON_RUN_AS_NODE'] || process.versions['electron'])) {
              return; // only on Electron / Electron-as-node
```

Reducing the new condition, `as unknown` is a TypeScript assertion is erased at compile time, contributing nothing at runtime -- so the guard is `Module.register && isElectron`. `Module.register` is a function on any Node >= 20.6 (I confirmed typeof = "function" on the bundled Node 24), so it's permanently truthy and the whole thing collapses to:

```js
if (isElectron) return;   // upstream said: if (!isElectron) return;
```

So exactly inverted. The retained comment `// only on Electron / Electron-as-node` now says the opposite of what the code does, which hints that this was a slip rather than intended behavioural change. The desktop app therefore never registers the resolve hook.

I found a few commits touching that area on the upstream ::

8 Jul | [913d5cccc1a](https://github.com/microsoft/vscode/commit/913d5cccc1a0e97311521e7791033ce9d5c2ef2f) `feat: restore asar for node_modules` ([#324897](https://github.com/microsoft/vscode/pull/324897))
8 Jul | [926982696e5](https://github.com/microsoft/vscode/commit/926982696e5076805ed790713fe2417a6eb2c7c2) `Revert "feat: restore asar for node_modules"` ([#324875](https://github.com/microsoft/vscode/pull/324875))
7 Jul | [c9dda3b36af](https://github.com/microsoft/vscode/commit/c9dda3b36af0c2ccd7e7692500910421dc683c67) `feat: restore asar for node_modules` ([#324084](https://github.com/microsoft/vscode/pull/324084))

so that was all in 1.129 release cycle.

---

I discovered this in a long debugging session when I rebased onto the today's (27 Aug) commit of VSCodium; at first I suspected that my `product.json` patcher is to blame, but it turns out that this patch is what causes the exthost to die.

`console.log` from a pristine-clean install with `--transient` :: <https://termbin.com/ma7a>.

After changing the patch to ::

```patch
❯ cat src/vscodium/patches/linux/00-remote-fix-bootstrap.patch
diff --git a/src/bootstrap-esm.ts b/src/bootstrap-esm.ts
index 49098a31..ae041b7d 100644
--- a/src/bootstrap-esm.ts
+++ b/src/bootstrap-esm.ts
@@ -6,3 +6,3 @@
 import * as fs from 'node:fs';
-import { register } from 'node:module';
+import * as Module from 'node:module';
 import { dirname, join } from 'node:path';
@@ -40,3 +40,3 @@ globalThis._VSCODE_FILE_ROOT = import.meta.dirname;
 function enableASARSupport(): void {
-       if (!process.env['ELECTRON_RUN_AS_NODE'] && !process.versions['electron']) {
+       if (Module.register as unknown && !(process.env['ELECTRON_RUN_AS_NODE'] || process.versions['electron'])) {
                return; // only on Electron / Electron-as-node
@@ -235,3 +235,3 @@ function enableASARSupport(): void {
 
-       register(`data:text/javascript;base64,${Buffer.from(jsCode).toString('base64')}`, import.meta.url, {
+       Module.register(`data:text/javascript;base64,${Buffer.from(jsCode).toString('base64')}`, import.meta.url, {
                data: process.env['VSCODE_DEV'] ? {} : {
```

, i.e. the entire change is ::

```diff
-+    if (Module.register as unknown && (process.env['ELECTRON_RUN_AS_NODE'] || process.versions['electron'])) {
++    if (Module.register as unknown && !(process.env['ELECTRON_RUN_AS_NODE'] || process.versions['electron'])) {
```

, I get the regular ::

```log
workbench.desktop.main.js:38  INFO Started local extension host with pid 79112.
workbench.desktop.main.js:38  INFO Started initializing default profile extensions in extensions installation folder. file:///tmp/vscode-HpJGtpXS/extensions
workbench.desktop.main.js:38  INFO Completed initializing default profile extensions in extensions installation folder. file:///tmp/vscode-HpJGtpXS/extensions
workbench.desktop.main.js:38  INFO [AccountPolicyGate] apply: state=inactive, reason=undefined, isRestricted=false
workbench.desktop.main.js:38  INFO [perf] Render performance baseline is 13ms
workbench.desktop.main.js:38  INFO [AccountPolicyGate] apply: state=inactive, reason=undefined, isRestricted=false
```

in the VSC DevTools (again, on an empty profile, `--transient`)


daiyam:
Ah! The condition has been inverted. Thanks for the catch.

daiyam:
Hum, after review, the patch need to be removed since the whole file has changed quite a bit. The original issue (https://github.com/VSCodium/vscodium/issues/2049) will need to be retested to see if it's still existing or not.

Can you update your PR? Thx

0zitro:
> the patch need to be removed since the whole file has changed quite a bit.

I hope I didn't misunderstand the request and you were talking about the patch as in "VSCodium patch in `patches/`" and not as in "the change this PR suggests" :)

PR #2984, opened 2026-08-30
fix(build): use configured organization in Electron metadata

ThadHouse:
Use ORG_NAME for Electron company and copyright resource substitutions. This preserves VSCodium's default metadata while allowing downstream builds to keep their main executable metadata consistent with other native binaries.

This PR was done with the help of AI, but was fully reviewed by me.


daiyam:
You will need to patch `prepare_vscode.sh` with your own urls, UUIDs, ... so I prefer to leave those.

Here how I do a custom build: https://github.com/zokugun/MrCode/blob/master/prepare.sh

